PR #60, opened 2019-11-28
Dockerize in prep for k8s-deploy

hedvigoscar:


sampettersson:
Do we have an NPM token in travis? If so could we auto publish the package while we are at it? 😍

hedvigoscar:
No, but we can easily add one 🙂

hedvigoscar:
The big question is when to publish the package and what to name the version 😉

sampettersson:
@hedvigoscar let's leave it for now then 😄

PR #83, opened 2019-12-03
Multiple expression support

hedvigoscar:
I'm sure we're approaching Turing completeness real soon here


hedvigoscar:
Optimizing compiler, here we come

PR #89, opened 2019-12-04
Fix various issues with MultiAction

hedvigoscar:
- `removeValues()` in KeyValueStore now only removes the requested
values, not everything in the store
- The `Label`-functionality in now actually works


hedvigoscar:
Sam forgot to approve it in github 😬

PR #124, opened 2019-12-09
Skip redirect-only passages in history

hedvigoscar:


src/Utils/useEmbark.ts
```diff
@@ -12,6 +12,26 @@ type Action =
...
+const shouldBeAddedToHistory = (passage: any) => {
+  if (passage.api) {
+    return false;
+  }
+
+  if (passage.externalRedirect) {
+    return false;
+  }
+
+  if (
```
sampettersson: haha remove this so we don't mislead people into thinking this is where this happen 😄

PR #128, opened 2019-12-10
Enhancements to switcher module

sampettersson:


src/Utils/useMeasure.ts
```diff
@@ -0,0 +1,27 @@
...
+import ResizeObserver from "resize-observer-polyfill";
+
+interface Measured {
+  left: number;
+  top: number;
+  width: number;
+  height: number;
+}
+
+export function useMeasure<T extends Element>() {
```
hedvigoscar:
```suggestion
export const useMeasure = <T extends Element>(): [{ ref: React.MutableRefObject<T> }, Measured] => {
```

javascript functions are weird and ma

PR #134, opened 2019-12-10
Regex again. Dammit!

hedvigoscar:


src/Components/Actions/MultiAction/util.test.ts
```diff
@@ -0,0 +1,18 @@
+import { getMultiActionItems } from "./util";
+
+describe("getMultiActionItems()", () => {
```
palmenhq: this is very minor but i think we should use the same style of writing tests as in the web onboarding. (using `it('blablabla, () => ...)` directly instead of `describe/it`

palmenhq: amazing with tests though!

src/Components/Actions/MultiAction/util.ts
```diff
@@ -0,0 +1,32 @@
...
+        return acc;
+      }
+
+      const index = matches[1];
+      const dataKey = matches[2];
+
+      return {
+        ...acc,
+        [index]: {
+          ...(acc[index] ? acc[index] : {}),
```
palmenhq:
```js
...(acc[index] ?? {})
```

hedvigoscar: My editor refuses 😬 I agree but yeah

hedvigoscar:
I agree - this function is certainly problematic. Lets try to replace it whenever the push is over

PR #147, opened 2019-12-12
Switcher onboarding

sampettersson:


src/Components/Icons/BankID.tsx
```diff
@@ -0,0 +1,184 @@
+import * as React from "react";
+
+export const BankID = () => (
+  <svg
```
hedvigoscar: Can we run this SVG through `svgo` or https://jakearchibald.github.io/svgomg/ or something similar? I feel like there's a lot of crap in this that is just not used 🤔

src/Components/Icons/DummyQRCode.tsx
```diff
@@ -0,0 +1,244 @@
+import * as React from "react";
+
+export const DummyQRCode = () => (
+  <svg
```
hedvigoscar: 💯

src/Components/Actions/ExternalInsuranceProviderAction/index.tsx
```diff
@@ -64,25 +77,41 @@ export const ExternalInsuranceProviderAction: React.FC<ExternalInsuranceProvider
   passageName,
   next
 }) => {
+  const { updateOperation } = React.useContext(BackgroundFetchContext);
   const { setValue } = React.useContext(StoreContext);
   const [state, setState] = React.useState({
+    id: uuid(),
```
palmenhq: should we really update this on each render?

PR #149, opened 2020-01-07
Add header logo escape link

palmenhq:


palmenhq:
actually i realized something @sampettersson ... What do we do with english?

sampettersson:
@palmenhq ah, true, add a field to Angel maybe?

palmenhq:
@sampettersson angel? Or just a prop from the embark component we use in web onboarding?

sampettersson:
@palmenhq I was thinking that different stories might have different exit paths, i.e making a claim: go back there, signing up: go back there, filling out a form: go back there.

sampettersson:
but do whatever you think is best :)

palmenhq:
@sampettersson hm, maybe. Sounds unlikely for a while though, i always expect the logo to take me to the start page. I guess the only case this might be true is if we add another "portal" or something that feels like a different site

sampettersson:
@palmenhq if we do a fully custom partner flow, for example, I would imagine that it might go back to the partner and not our website. It's nice for editors to have the flexibility to do whatever they please without involving us hehe 😄

sampettersson:
@palmenhq you could add it here quite easily: https://github.com/HedvigInsurance/Angel/blob/c27f2fa57e6e8462608cebb8985a8d7aa5255eb6/src/dialogs/partner-configs/index.html#L6

and then it will just be in the payload under partnerConfigs, it's what's currently deciding which logo we show, so I think it makes sense?

palmenhq:
@sampettersson sure, i'll push this for later though because this grew in scope quite a lot i think. I thought it was just making a link 😂

palmenhq:
@sampettersson i think this solution is better, leaving the escape link to whoever consumes the header (currently the web onboarding). Thoughts?

PR #171, opened 2020-01-13
GraphQL

sampettersson:
I've probably missed something somewhere, but gonna test this out for reals and then add it if we miss something later.


schema.ts
```diff
@@ -0,0 +1,279 @@
...
+`;
+
+export const schema = makeExecutableSchema({
+  typeDefs,
+  resolvers: {
+    Query: {
+      angelStory: async (_, { name }) => {
+        if (name.includes("../")) {
+          throw new Error("Can't traverse downwards");
+        }
```
palmenhq: clever! i think though we should maybe do a more safe check. Maybe we could read all file names in `angel-data` and just check if `name` is in that array, and if that's the case read the file? This feels a little shakey

src/parseStoryData.ts
```diff
@@ -744,17 +771,44 @@ export const parseStoryData = (storyData: any) => ({
...
+        const toAttribute = redirect.getAttribute("to");
+        const links = parseLinks(toAttribute || "");
+
+        const expression = parseExpression(whenAttribute || "");
+
+        if (!expression || !links) {
+          return null;
+        }
+
+        console.log(expression.__typename);
```
palmenhq: 🙈

PR #194, opened 2020-02-27
Add support for generic GraphQL

sampettersson:
So right now each time you want to do some API integration we have to create custom implementations, this is boring and sad to do, so I took the liberty to do a generic implementation.

you use it like this:

```html
<GraphQLApi next="[[correctStreetAddressApartment]]">
    <Query>
      query PersonalInformation($personalNumber: String!) {
			personalInformation(input: { personalNumber: $personalNumber }) {
				streetAddress
	  		}
	  }
    </Query>
    <Variable key="personalNumber"
              from="personalNumber"
              as="string"
    >
    </Variable>
    <Error next="[[manualInfoVilla]]">
    </Error>
    <Result key="personalInformation.streetAddress"
            as="streetAddress"
    >
    </Result>
</GraphQLApi>
````

Also parcel hated graphql, so I switched it out to webpack.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,14 +15,12 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
+    "@babel/plugin-proposal-optional-chaining": "^7.8.3",
     "@emotion/core": "^10.0.21",
     "@emotion/styled": "^10.0.17",
     "@hedviginsurance/brand": "^2.0.0",
-    "@types/lodash.get": "^4.4.6",
     "animated-scroll-to": "^2.0.5",
     "apollo-link-http": "^1.5.16",
-    "awesome-typescript-loader": "^5.2.1",
-    "babel-loader": "^8.0.6",
     "date-fns": "^2.8.1",
     "emotion": "^10.0.17",
     "eventemitter3": "^4.0.0",
@@ -67,10 +65,13 @@
     "react-dom": "^16.10.x"
   },
   "devDependencies": {
+    "@babel/preset-react": "^7.8.3",
+    "@babel/preset-typescript": "^7.8.3",
     "@koa/cors": "^2.2.3",
     "@types/history": "^4.7.3",
     "@types/jest": "^24.0.23",
     "@types/koa": "^2.0.51",
+    "@types/lodash.get": "^4.4.6",
     "@types/react": "^16.9.9",
     "@types/react-dom": "^16.9.2",
     "@types/react-input-mask": "^2.0.4",
@@ -78,7 +79,6 @@
     "concurrently": "^5.0.0",
     "graphql-tag": "^2.10.1",
     "jest": "^24.9.0",
-    "parcel-bundler": "^1.12.4",
     "prettier": "^1.19.1",
     "rollup": "^1.27.2",
     "rollup-plugin-typescript2": "^0.25.2",
@@ -87,7 +87,8 @@
     "webpack-cli": "^3.3.11",
     "webpack-dev-middleware": "^3.7.2",
     "webpack-dev-server": "^3.10.3",
-    "webpack-merge": "^4.2.2"
+    "webpack-merge": "^4.2.2",
+    "babel-loader": "^8.0.6"
   },
   "resolutions": {
     "graphql": "^14.6.0"
--- a/package.json
+++ b/package.json
@@ -8,8 +8,7 @@
     "test": "jest",
     "server": "ts-node --project tsconfig.server.json server.ts",
     "build-package": "rollup -c",
-    "build": "parcel build src/index.tsx --out-dir dist/output",
-    "watch": "parcel watch src/index.tsx --out-dir dist/output",
+    "watch": "webpack-dev-server",
     "dev": "concurrently -n \"server, watch\" \"npm run server\" \"npm run watch\"",
     "prepublishOnly": "rollup -c"
   },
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

**Comment by hedvigoscar** [comment]
@palmenhq 

I'm not entirely comfortable with this on the basis that it could open us up to more GraphQL calls being made than we're ok with through the angel tree. What are your thoughts on this?

**Comment by sampettersson** [comment]
Me and @hedvigoscar discussed this and thought it would be best to add a request filter in giraffe which looks at the data it should return and see if the field has a custom `@embark` directive that allows embark to fetch it, if it doesn't giraffe would fail the request. 

In that way we have better control over what embark is allowed to do :)

**Comment by palmenhq** [comment]
@sampettersson @hedvigoscar I don't understand the issue here, is it a security thing? Or are we afraid of the angel authors abusing it?

**Comment by palmenhq** [comment]
Having discussed this with @hedvigoscar irl now, i don't agree this is a way to move forward. I think its better to fix the auth solution properly at some point, instead of fixing it by just making a potential attack harder to carry out. Its better to not fix it, and once it gets important to patch this (i.e. when we introduce "dangerous" mutations or expose sensitive data via graphql) we patch it properly. Thoughts?

File: package.json [hunk 1]
```diff
@@ -34,20 +38,21 @@
     "koa-graphql": "^0.8.0",
     "koa-mount": "^4.0.0",
     "koa-static": "^5.0.0",
+    "lodash.get": "^4.4.2",
+    "parcel": "^2.0.0-alpha.3.2",
```
palmenhq: if we're removing parcel, shall we remove it from here also?

package.json
```diff
@@ -19,12 +18,17 @@
     "@emotion/core": "^10.0.21",
     "@emotion/styled": "^10.0.17",
     "@hedviginsurance/brand": "^2.0.0",
+    "@types/lodash.get": "^4.4.6",
```
palmenhq: `devDependencies` :)

webpack.config.js
```diff
@@ -0,0 +1,74 @@
...
+  module: {
+    rules: [
+      {
+        test: /\.js$/,
+        use: ["babel-loader", "source-map-loader"],
+        exclude: /node_modules/
+      },
+      {
+        test: /\.tsx?$/,
+        use: ["babel-loader", "awesome-typescript-loader"]
```
palmenhq: why do we need `atl` here? Babel can transpile typsecript since https://babeljs.io/docs/en/babel-preset-typescript and thats what we use in all other projects :)

src/Components/API/graphql.ts
```diff
@@ -0,0 +1,4 @@
+import { GraphQLApiComponent, ApiComponent } from "./apiComponent";
+
+export const isGraphqlApi = (t?: ApiComponent): t is GraphQLApiComponent =>
+  (t && t.component === "GraphQLApi") || false;
```
palmenhq: `t?.component === 'GraphqlApi`?

src/Components/API/index.ts
```diff
@@ -21,6 +24,80 @@ export const callApi = async (
...
+    try {
+      const variables = component.data.variables.reduce((curr, variable) => {
+        switch (variable.as) {
+          case "string":
+            return { ...curr, [variable.key]: String(store[variable.from]) };
+          case "int":
+            return { ...curr, [variable.key]: parseInt(store[variable.from]) };
+        }
+
+        return { ...curr, [variable.key]: store[variable.from] };
```
palmenhq: maybe we should put this as a default branch in the switch statement? Just to make it a bit easier to read :)

src/Components/API/index.ts
```diff
@@ -21,6 +24,80 @@ export const callApi = async (
...
+        );
+
+        if (graphqlMutationResult.data) {
+          handleData(graphqlMutationResult);
+        } else if (graphqlMutationResult.errors) {
+          handleErrors(graphqlMutationResult);
+        }
+
+        return;
+      }
```
palmenhq: i feel like we're doing almost the exact same logic twice here, can we maybe refactor it a bit more neat?

src/Components/Actions/NumberActionSet/NumberEditCard.tsx
```diff
@@ -3,8 +3,7 @@ import { Tooltip } from "../../Tooltip";
 import { motion } from "framer-motion";
 import { InlineNumberAction } from "./InlineNumberAction";
 import styled from "@emotion/styled";
-import { colorsV2 } from "@hedviginsurance/brand/colors";
-import { fonts } from "@hedviginsurance/brand/fonts/index";
```
palmenhq: ❤️

src/parseStoryData.ts
```diff
@@ -641,6 +639,104 @@ const getFirstLevelNodes = (node: Element) => {
...
+
+    const parseResults = (element: Element) => {
+      const key = element.getAttribute("key");
+      const as = element.getAttribute("as");
+
+      return {
+        key,
+        as
+      };
+    };
```
palmenhq: Maybe we could move these helper functions out? To me it seems they do not _have_ to be within this function and it'd be easier to grasp what `parseApi` does if there's no clutter inside it

src/parseStoryData.ts
```diff
@@ -641,6 +639,104 @@ const getFirstLevelNodes = (node: Element) => {
...
+      const nextLinks = parseLinks(graphQLApi.getAttribute("next") || "");
+      const variables = Array.from(
+        graphQLApi.getElementsByTagName("variable")
+      ).map(parseVariables);
+      const errors = Array.from(graphQLApi.getElementsByTagName("error")).map(
+        parseErrors
+      );
+      const results = Array.from(graphQLApi.getElementsByTagName("result")).map(
+        parseResults
+      );
```
palmenhq: clean 😍

src/parseStoryData.ts
```diff
@@ -641,6 +639,104 @@ const getFirstLevelNodes = (node: Element) => {
...
+    const variables = Array.from(
+      graphQLApi.getElementsByTagName("variable")
+    ).map(parseVariables);
+    const errors = Array.from(graphQLApi.getElementsByTagName("error")).map(
+      parseErrors
+    );
+    const results = Array.from(graphQLApi.getElementsByTagName("result")).map(
+      parseResults
+    );
+
```
palmenhq: but same here, could we find a more clever way to not duplicate the query/mutation logic?

sampettersson:
@palmenhq updated now, didn't really figure out a good way to split `query` & `mutation` so didn't do that but the logic into separate files so that it's less messy atleast 😇

PR #214, opened 2020-03-24
Update embark flow to handle norwegian data

vonElfvin:
- Add key value ability to redirect
- Add mask to norwegian postal codes
- Add mask to birthdate with ability to extract age

### DIFF
--- a/schema.ts
+++ b/schema.ts
@@ -311,19 +311,25 @@ const typeDefs = `
     type EmbarkRedirectUnaryExpression {
         type: EmbarkExpressionTypeUnary!
         to: String!
+        passedExpressionKey: String!
+        passedExpressionValue: String!
     }
 
     type EmbarkRedirectBinaryExpression {
         type: EmbarkExpressionTypeBinary!
+        to: String!
         key: String!
         value: String!
-        to: String!
+        passedExpressionKey: String!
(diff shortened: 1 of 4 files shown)


sampettersson:
FYI You also need to add these to the GraphQL schema!

src/Utils/ExpressionsUtil/useGoTo.ts
```diff
@@ -31,10 +32,11 @@ export const useGoTo = (
...
 
         if (passableExpressions.length > 0) {
-          const { to } = passableExpressions[0];
-          const redirectTo = data.passages.filter(
+          const { to, key, value } = passableExpressions[0];
+          const redirectTo = data.passages.find(
             (passage: any) => passage.name == to
-          )[0];
+          );
+          setValue(key, value)
```
sampettersson: check if key and value is null before doing a setValue

schema.ts
```diff
@@ -311,19 +311,25 @@ const typeDefs = `
     type EmbarkRedirectUnaryExpression {
         type: EmbarkExpressionTypeUnary!
         to: String!
+        passedExpressionKey: String!
```
sampettersson: These can be null right? So just ”String”.

PR #231, opened 2020-03-31
Update to Web Onboarding NO - English Travel

zakkjellin:


vonElfvin:
Looks like this is updating the english one rather than creating a new Norwegian one?

palmenhq:
Seems stale, i'll close this. lemme know if we want it back

PR #289, opened 2020-05-08
Offer keys and values as lists instead of strings which have to be in…

hedvigoscar:
…terpreted by consumers


hedvigoscar:
I mostly wanted to affect the Embark GraphQL, but I did my best to provide full backwards compatibility for web.

PR #293, opened 2020-05-13
add lan stockholm and aktsam to providers

AlexSJ1990:
used the same logo for Lans Stockholm as Lans and used the logo for Aktsam that is in content service


src/Components/Actions/ExternalInsuranceProviderAction/providers.tsx
```diff
@@ -543,5 +558,33 @@ export const norwegianProviders: ReadonlyArray<Provider> = [
...
       </svg>
     ))
+  },
+  {
+    id: "Aktsam",
+    externalCollectionId: "AKTSAM",
+    name: "Aktsam",
+    hasExternalCapabilities: true,
+    icon: withIcon(30)(() => (
+      <svg width="298" height="87" xmlns="http://www.w3.org/2000/svg">
```
palmenhq, line 568: Don't we need viewBox here? Or maybe it works without i'm not sure

PR #300, opened 2020-05-25
Fix GraphQL API mistakes

sampettersson:
Mistakes where made


palmenhq:
Hello, this broke backwards compatibility so web onboarding is broken in staging atm. Could we make a fix that is backwards compatible plz?

PR #305, opened 2020-07-07
Align this project with other projects

palmenhq:
Its a lot changes but its mostly because i ran prettier on all the files. Key points:

- Use same prettier configuration
- Use yarn instead of npm
- Align the commands with other projects (`yarn watch` to start a dev server, for example)


src/Components/Actions/MultiAction/util.ts
```diff
@@ -1,32 +1,32 @@
...
+      const index = matches[1]
+      const dataKey = matches[2]
 
       return {
         ...acc,
         [index]: {
           ...(acc[index] ? acc[index] : {}),
-          [dataKey]: store[storeKey]
-        }
-      };
```
rasmusguterstam, line 31: Are you sure about this one? 🧐

palmenhq: it just some semis :)

PR #318, opened 2020-08-03
Add query for getting all names of stories

sampettersson:


schema.ts
```diff
@@ -6,6 +6,8 @@ import { parseStoryData } from "./src/Parsing/parseStoryData";
 const typeDefs = `
     type Query {
         embarkStory(name: String!): EmbarkStory
+        # returns names of all available embark stories
+        embarkStories: [String]
```
hedvigoscar, line 10:
```suggestion
        embarkStoryNames: [String!]!
```

Another idea could be to have:

```graphql
embarkStories: [EmbarkStoryThing!]!

# Not the greatest name but you get the idea
type EmbarkStoryThing {
  name: String!
  story: EmbarkStory! # This could be resolved lazily in order to not incur a penalty each time we query this type
}
`

PR #320, opened 2020-08-05
Insurance providers through graphql

sampettersson:


schema.ts
```diff
@@ -2,12 +2,45 @@ import { storyKeywords } from "./src/storyKeywords";
...
 
 const typeDefs = `
     type Query {
         embarkStory(name: String!): EmbarkStory
         # returns names of all available embark stories
         embarkStoryNames: [String!]!
+        embarkInsuranceProviders(country: EmbarkCountry!): [EmbarkInsuranceProvider!]!
+    }
+
+    type Icon {
```
hedvigoscar, line 15: Hmm. We have basically all of the infrastructure/schema for this available in `content-service`, right? I think we should probably express this as a cross-schema resolution in that case.

PR #321, opened 2020-08-06
Fix/boolean bug

gustaveen:
When adding an extra building and setting `hasWaterConnected` to `true`, it would actually be passed to the backend as `false`


gustaveen:
🙌

PR #322, opened 2020-08-17
Resolve birthdate reverse age correctly

palmenhq:
Since `BirthDateReverse` is "unmasked" before its passed to `derivedValues` it is always in iso format when it arrives there.

### DIFF
--- a/src/Components/Actions/masking.test.tsx
+++ b/src/Components/Actions/masking.test.tsx
@@ -1,0 +1,127 @@
+import {
+  derivedValues,
+  isValid,
+  mapMaskedValue,
+  mapUnmaskedValue,
+  unmaskValue,
+} from './masking'
+import { differenceInYears } from 'date-fns'
+
+describe('isValid', () => {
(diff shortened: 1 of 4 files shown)


palmenhq:
@gustaveen YES, excellent push. Bakläxa for me!

src/Components/Actions/masking.test.tsx
```diff
@@ -0,0 +1,127 @@
...
+      '1912-12-12',
+    )
+  })
+
+  it("doesn't map invalid reverse birth date", () => {
+    expect(mapUnmaskedValue('1234567', 'BirthDateReverse')).toBe('1234567')
+  })
+
+  it("doesn't map other masks", () => {
+    expect(mapUnmaskedValue('1912-12-12', 'BirthDate')).toBe('1912-12-12')
```
gustaveen, line 90: Should we maybe use a data where day and month isn't the same :) 1912-12-30?

palmenhq: yes, good point haha!

PR #326, opened 2020-08-27
Unscew logo

gustaveen:
The logo was a bit squashed, caused by margin-top


palmenhq (approved):
not sure i understand this but if it works it works i guess 🤷‍♀️

PR #357, opened 2020-10-26
Update to Web Onboarding - Swedish Switcher

zakkjellin:
Small copy change that Konsumentombudsmannen has required us to do before october 31st, plz merge before then 😂


gustaveen (approved):
KO!
ko

PR #361, opened 2020-11-11
DK Flow

palmenhq:
including graphql mutation 🥳  <strike>that may or may not work, we'll see. But i think it should be ok</strike>


vonElfvin:
The "test flow" in angel logs out if the mutation is valid given the schema if I remember correctly 😌

palmenhq:
@vonElfvin clever! It had some csp issues since we migrated to "online" but i tested and it actually seems to work 😮

vonElfvin (approved):
LGTM 👍

PR #365, opened 2020-11-18
Computed Store values

palmenhq:
Allows for using "computed" store values anywhere in embark. They are administered via Angel and can be use to, for example, send `coInsured` to backend while gathering `householdSize` to the store.

See HedvigInsurance/embark-docs#7 for details on usage.


sampettersson:
I actually prefer us implementing templating on our own as if we were to implement handlebars or something similar we would have an enormous spec to implement natively on iOS and Android, that is kind of the core reason for why we kept templating so barebones initially.

But speaking of adding specifically these features I think they make sense :)

Now when this feature is becoming so complex maybe it makes sense to start writing down an official specification for this? We would need to implement this in the apps in the coming weeks.

Another solution to this problem could be to add a component that computes values from the result of an expression? Something like:
 `<ComputedValue name="coInsured" expression="houseHoldSize - 1" />`

Then you only have to have that logic once and not in every place that needs `coInsured`, and it would work in when expressions automatically as you would just use `coInsured` directly instead.

hedvigoscar:
I agree with Sam, we need a good spec in place in order to be able to replicate this functionality in the apps. `<ComputedValue>`s could also be pretty neat, I think!

sampettersson:
I also think `ComputedValues` can remove some complexity in places where it's already introduced, like calculating age from a personal number.

Basically, the store could work something like this instead:

```javascript
const store = {}

store.householdSize = (_) => "2"
store.coInsured = (store) => store.houseHoldSize() -1

const Component = () =>
  <h1>{store.coInsured()}</h1>
```

palmenhq:
Good points! I like having xml tags since everything else is xml anyways but this means we need to start parsing xml in the clients too right? I like the idea of computed values but i'm not sure i agree computed values would be the solution to the problem i'm solving myself rn. Shall we have a quick discussion around this? Feels like we're discussing several issues in parallel now.

palmenhq:
@sampettersson @hedvigoscar can u check again plox?

sampettersson (approved):

hedvigoscar (commented):
Other than the reservations I have wrt. string concatenation, LGTM!

gustaveen (approved):
🔤 🔢 👌

PR #369, opened 2020-11-25
Added embarkStories query

FredrikAreschoug:
```query { embarkStories(locale) } ``` will resolve metadata for the stories on locale. 

The metadata is included in the story json so please have a lookout so they don't get removed in any upcoming pr :slightly_smiling_face: . I have tested and the metadata is not overridden by angel. But the will be removed if you use old stories, so I will talk to zak about that.

### CODE REVIEW

**Comment by hedvigoscar** [comment]
LGTM!

Would be nice to get the metadata into the individual files, but it's not super-pressing I guess.

Sidenote: Are we doing Disk I/O for each call into the embark-service? It would probably be not that hard to load the files into memory on startup, right? Could potentially save a lot of time for our users 😬

File: schema.ts [hunk 1]
```diff
@@ -420,6 +422,25 @@ const typeDefs = `
...
+    union EmbarkStoryMetadataEntry = EmbarkStoryMetadataEntryDiscount
+
+    type EmbarkStoryMetadataEntryDiscount {
+      discount: String!
+    }
+
+    enum EmbarkStoryType {
+      WEB_ONBOARDING
+      APP_ONBOARDING
+    }
```
palmenhq, line 442: this stuff is brilliant!

src/Resolvers/resolveStoriesMetadata.ts
```diff
@@ -0,0 +1,17 @@
+import { promises } from 'fs'
+
+export const resolveMetadataOnLocale = async (locale: string) => {
+  switch (locale) {
+    case 'sv_SE': {
+      const file = await promises.readFile(
+        'stories-metadata/Swedish - Sweden - Stories.json',
+        {
```
palmenhq, line 8: commonly in node you should be careful with reading things from relative paths like this. Rather, we could use

```ts
await promises.readFile(
        path.resolve(__dirname, '../../stories-metadata/Swedish - Sweden - Stories.json'),
        {
// ...
```

to be sure this is always present relative to this scirpt file rather than where you run the script from :)

stories-metadata/Swedish - Sweden - Stories.json
```diff
@@ -0,0 +1,30 @@
+[
```
palmenhq, line 1: instead of splitting this into multiple files and mapping it with switch, what about having one big file which is a map with arrays like this?

something like 

```ts
{ "sv_SE": [
    { "title":  "Teckna blabla", "description": "Se pris bla bla", ... }
] }
```

then when resolving it we could just do something like

```ts
export const resolveMetadataOnLocale = async (locale: string) => {
  // ...
  // read file
  // ...
  return resolvedFile[locale]
}
```

FredrikAreschoug:
@hedvigoscar Cool! 
But what do you mean with `Would be nice to get the metadata into the individual files, but it's not super-pressing I guess.` do you think all objects in `Sweden - Swedish - Stories.json` should be slit up in different files?

On the side not that sounds awsome! but I don't feel that I have the skills to do that in a good way for this service, just went along with what we did in the other resolver :slightly_smiling_face:

schema.ts
```diff
@@ -3,12 +3,14 @@ import { storyKeywords } from './src/storyKeywords'
...
 import { promises } from 'fs'
 import { parseStoryData } from './src/Parsing/parseStoryData'
+import { resolveMetadataOnLocale } from './src/Resolvers/resolveStoriesMetadata'
 
 const typeDefs = `
     type Query {
         embarkStory(name: String!, locale: String!): EmbarkStory
         # returns names of all available embark stories
         embarkStoryNames: [String!]!
+        embarkStories(locale: String!): [EmbarkStoryMetadata!]!
```
palmenhq, line 13: Maybe this should be nullable? In case there's no such locale. With the current implementation apollo will give some cryptic error message right?

```ts
embarkStories(locale: String!): [EmbarkStoryMetadata!]
```

palmenhq (commented):
Looks awesome, some comments :) I think rather we should have them in one file in order to not complicate things more than needed... what's your thoughts?

Re the disk io thing - node is very performant with disk io and parsing json afaik so i _think_ we wouldn't save more than a few milliseconds probably (but i may of course be off here). On the other hand - since its so little data i don't think it'd hurt us memory-wise. But i think saving _a lot_ of time for the end user would be bold to say 😬

stories-metadata/Swedish - Sweden - Stories.json
FredrikAreschoug: Yeah true I like it it will be a lot easier to edit it from angel then! will changing to that!

schema.ts
FredrikAreschoug: I think we commonly use empty list instead of null, but whatever you like the most :slightly_smiling_face:

hedvigoscar:
I think in my optimal case the metadata would exist alongside the story - i.e `Web Onboarding - Swedish Switcher.json` has the metadata for itself.

I guess a lot of time is a matter of perspective - disk I/O is measured in milliseconds vs memory access which is measured in nanoseconds. On an individual request level it's not a huge amount of time, naturally, but collectively it would add up.

It's also worth noting that we're not gaining anything by reading from disk, other than the time it would take to implement pre-loading of this data into memory 😉

FredrikAreschoug:
True oscar maybe it makes more sense to add the metadata in the actual file and then we would add the locale in the store like 
`Web Onboarding - Swedish Switcher.json`: {
  "id": "892533f3-934e-40ac-bccf-238e15f634e2",
  "ifid": "2F54B9F2-B6C1-4CFC-BF17-4DB41DCB1ED0",
  "lastUpdate": "2020-11-11T15:32:51.140Z",
  "name": "Web Onboarding DK - Contents",
  "locale": "sv_SE"
  "passages": [...]
  "metadata": [
    {
      "title": "Teckna ny hemförsäkring",
      "description": "Se pris och skaffa Hedvig",
      "type": "WEB_ONBOARDING",
      "metadata": []
    },
    {
      "title": "Teckna ny hemförsäkring",
      "description": "Se pris och skaffa Hedvig",
      "type": "APP_ONBOARDING",
      "metadata": []
    }
  ]
}


Is that better pls let me know and I will do the change. Another question will I have to fix Angel then, so the metadata is not overridden each change?

hedvigoscar:
This will be overwritten by Angel, yes 😧

FredrikAreschoug:
Looking into angel I don't think I will manage to fix this anytime soon so I will open this up for review as is

schema.ts
palmenhq: Normally i'd agree, but i think there's a difference between having 0 metadata entrie in the locale and not having a locale. Or? :)

palmenhq:
Good points! One thought that struck me now, shall we maybe resolve text keys directly inside here instead of hard-coding it in the files?

FredrikAreschoug:
Are we resolving text keys in other places in this project? If we do please point it out, and I can change and we should go on location/market and not locale. If we don't I would say that's a scope creep I don't want to take that on

palmenhq:
We do! checking [schema.ts](https://github.com/HedvigInsurance/embark/blob/master/schema.ts#L444-L449) and [parseStoryData.ts](https://github.com/HedvigInsurance/embark/blob/f94b894dea92734d0dffbd35ddd349e932ee22d4/src/Parsing/parseStoryData.ts#L841-L845) should be helpful i think :) maybe we can do it differently but imo it would make sense

FredrikAreschoug:
@palmenhq looking into the text keys, I feel like there should probably break out some stuff from `parseStoryData.ts` like `createParsePossibleExpressionContent` for example. And then I saw that there is no json files for Norway at https://translations.hedvig.com/embark/ and I don't even know how to edit that. With that I just feel like this is growing in scope. But do you feel this or just create depth then let's close it

.idea/embark.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
palmenhq, line 1: perhaps we should add `.idea` to gitignore :)

schema.ts
```diff
@@ -451,6 +472,10 @@ export const schema = makeExecutableSchema({
 
         return storyData
       },
+      embarkStories: async (_, { locale }: { locale: string }) => {
+        const metadata = resolveMetadataOnLocale(locale)
```
palmenhq, line 476: maybe we should `await` this?

palmenhq: It'll still work though since `async`/`await` is just sugar for `Promise` and `Proimse`s automatically flatmap

src/Resolvers/resolveStoriesMetadata.ts
```diff
@@ -0,0 +1,40 @@
+import { promises } from 'fs'
+
+const path = require('path')
+
+export const resolveMetadataOnLocale = async (locale: string) => {
```
palmenhq, line 5: Could we perhaps put a return type here? Curretly this'll implicitly be `any`.

src/Resolvers/resolveStoriesMetadata.ts
```diff
@@ -0,0 +1,40 @@
...
+      const file = await promises.readFile(
+        path.resolve(__dirname, `../../angel-data/${name}`),
+        {
+          encoding: 'utf-8',
+        },
+      )
+
+      const json = JSON.parse(file)
+
+      if (json['locales'] && json['locales'].includes(locale)) {
```
palmenhq, line 24: you could also use `json.locales?.includes(locale)` which imo is a bit more pretty :)

src/Resolvers/resolveStoriesMetadata.ts
```diff
@@ -0,0 +1,40 @@
...
+        return json['metadata'].map((metadata: any) => {
```
palmenhq, line 25: maybe this should rather be a `Metadata` type?

palmenhq (commented):
really neat solution! Just some ts semantics

palmenhq (commented):
you will also ask git to stop tracking the `.idea` directory. You can do it with `git rm --cached .idea`

FredrikAreschoug:
Only did it for one file :man_facepalming:

palmenhq (approved):
awesome!

PR #370, opened 2020-11-27
Add fixed node version

palmenhq:


package.json
```diff
@@ -14,6 +14,9 @@
     "watch": "concurrently -n \"start, watch-client\" \"npm run start\" \"npm run watch-client\"",
     "prepublishOnly": "rollup -c"
   },
+  "engines": {
+    "node": "^12.19.0"
```
styrelius, line 18: Shouldn't it be
```suggestion
    "node": "12.19.0"
```
? Or maybe that doesn't really matter if we specify the. version in the `.nvmrc`? 🤔

palmenhq: In that case i guess we should match `.nvmrc` exactly, but imo it's not that important to match the minor/patch version exactly, as long as its node 12, which i guess is what this'd indicate

gustaveen (dismissed):

gustaveen (approved):
🔐 🙌

PR #377, opened 2020-12-16
removed unwanted slash

MathiasOhrstrom:


palmenhq (changes requested):
I don't agree with this, imo these should be treated as absolute paths and absolute paths should start with `/`. what's the reasoning behind this? :)

MathiasOhrstrom:
> I don't agree with this, imo these should be treated as absolute paths and absolute paths should start with `/`. what's the reasoning behind this? :)

Its only that the base url we have in the app ends with `/` so its no biggie if you think its better with the `/` in the beginning. But why do you think it should be the way you propose?

palmenhq (commented):
@MathiasOhrstrom right! I think it's more conventional to have base urls without trailing slash and have anything appended to that with leading slash (or at least in the web and backend world, i have no clue on coventions in the app universe), in case you for some reason would like to do for example an equality check on urls manually a trailing slash in the base url + the base path being `/` can cause issues from my experience.

I realize these changes are very much app specific so maybe this is ok, assuming that this would be the preferred way also for @HedvigInsurance/ios ?

MathiasOhrstrom:
@palmenhq Thanks for the explanation! That makes sense. I have changed the base URL in the app and @HedvigInsurance/ios have not started on this, so Ill close this PR

palmenhq:
@MathiasOhrstrom alrighty, sounds great :) if you could please delete the branch also that would be great so we don't get too many stale branches laying around. (You can always wake it from the dead if you would discover this is actually needed later)

PR #391, opened 2021-01-08
Add storybook to embark

gustaveen:
## What

Add storybook

## Why 
Easier component development

Also Storybook has updated the way to write stories a bit if it looks slightly different from before: https://storybook.js.org/docs/react/writing-stories/introduction#how-to-write-stories

https://user-images.githubusercontent.com/6661511/104033391-bf083b80-51cf-11eb-8ff7-3010e7ce2ade.mov

### DIFF
--- a/src/Components/BackButton/BackButton.stories.tsx
+++ b/src/Components/BackButton/BackButton.stories.tsx
@@ -1,0 +1,13 @@
+import React from 'react'
+import { Story, Meta } from '@storybook/react/types-6-0'
+import { BackButton, BackButtonProps } from './BackButton'
+
+export default {
+  title: 'BackButton',
+  component: BackButton,
+  argTypes: { onClick: { action: 'clicked' } },
+} as Meta
+
+const Template: Story<BackButtonProps> = (args) => <BackButton {...args} />
+
+export const Default = Template.bind({})
--- a/src/Components/Message/Message.stories.tsx
+++ b/src/Components/Message/Message.stories.tsx
@@ -1,0 +1,22 @@
+import React from 'react'
+import { Story, Meta } from '@storybook/react/types-6-0'
+import { Message, MessageProps } from './Message'
+
+export default {
+  title: 'Message',
+  component: Message,
+} as Meta
+
+const message = {
+  expressions: [],
+  text:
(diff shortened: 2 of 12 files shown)


src/Components/BackButton/BackButton.stories.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+export default {
+  title: 'BackButton',
+  component: BackButton,
+  argTypes: { onClick: { action: 'clicked' } },
+} as Meta
+
+const Template: Story<BackButtonProps> = (args) => <BackButton {...args} />
+
+export const Default = Template.bind({})
```
styrelius, line 13: Just curious since I don't understand - what does this `Template.bind({})` do? Why is it not possible to just export the `Template` (but name it `Default` I guess)?

src/Components/Message/Message.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+    'Okej! Då behöver vi bara ditt personnummer så kan vi automatiskt hämta resten av informationen som behövs för att ge dig ett pris',
+}
+
+const Template: Story<MessageProps> = (args) => <Message {...args} />
+
+export const Default = Template.bind({})
+Default.args = {
+  isResponse: false,
+  message: message,
+}
```
styrelius, line 22: I also don't understand this heh 🙈 Is it not possible to add these as default `args` on line 16 instead? Like 
```javascript
const Template: Story<MessageProps> = (args = { isResponse: false, message: message, }) => <Message {...args} />
```

styrelius (commented):
Really nice! Makes a ton of sense to have Storybook here 🙌 ✨  

Just two things I didn't understand 🤕

src/Components/BackButton/BackButton.stories.tsx
gustaveen: No worries! Storybook has updated the way you write stories a bit. The template thingy makes it easier to add variants to a story but also enables controls automagically from the `args` so you don't have to manually define them as before 🎉 The docs explains it better than I can 😃 https://storybook.js.org/docs/react/writing-stories/introduction#how-to-write-stories

styrelius: Ah OK 🆒 👌

styrelius (approved):
Nice! 👍

PR #392, opened 2021-01-13
Add story for TextAction

gustaveen:
## What 
Add stories for TextAction

https://user-images.githubusercontent.com/6661511/104480125-84821280-55c4-11eb-986c-aa7f7ce5f603.mov


src/Components/Actions/NumberAction/NumberAction.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+    name: 'personalNumberApartment',
+  },
+  mask: 'Email',
+  minValue: '1',
+  placeholder: '2',
+  storeKey: 'householdSize',
+  tooltip: {
+    title:
+      'Exempelvis partner, barn eller roomies som ska täckas. De medförsäkrade måste bo tillsammans med dig för att täckas av din försäkring',
+    description: 'Antal försäkrade',
```
styrelius, line 28: I only get the `description` when I hover the question mark, not the `title` - isn't the `title` text the one we want?

image

styrelius: And the size of that image is insane, sorry 😳

src/Components/Actions/TextAction/TextAction.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export const PersonalNumber = Template.bind({})
+PersonalNumber.args = {
+  isTransitioning: false,
+  passageName: 'personalNumberApartment',
+  mask: 'PersonalNumber',
+  storeKey: 'personalNumber',
+  placeholder: 'ååmmdd-xxxx',
+}
```
styrelius, line 34: Shouldn't the `PersonalNumber` be a `NumberAction`?

styrelius (commented):
Awesome! 🥳  Just two things 👇

src/Components/Actions/NumberAction/NumberAction.stories.tsx
gustaveen: I copied the tooltip content from what we have in production today, but I accidentally switched places on title and description 😄 (if that caused the confusion) However the title is only used in mobile where you get a Modal 👍 


https://user-images.githubusercontent.com/6661511/104610184-1d2b9780-5684-11eb-929a-d023da715d75.mov

src/Components/Actions/TextAction/TextAction.stories.tsx
gustaveen: We only use it for simple numbers like household size. For the same reason that we switched it in the ssn input in the checkout, you don't want to accidentally increase you personal number if pressing a number key for example :)

styrelius: AH, you are so right! 🧠

styrelius (approved):
NICE 😀

src/Components/Actions/NumberAction/NumberAction.stories.tsx
styrelius: 👍

src/Components/Actions/TextAction/TextAction.stories.tsx
gustaveen: I thought the exact same thing when making the stories so you're not alone 😄

PR #393, opened 2021-01-19
Enable external capabilities for new insurance companies

vonElfvin:


styrelius (approved):
Nice! This certainly seems like it should do the trick 👌 ✨

PR #395, opened 2021-01-21
Added sub type to metadata

FredrikAreschoug:
Got a request from Tarik to create an identifier for our stories I'm not sure if this is the best but let's have it as a start of a discussion on how to do this in the best future prof way :slightly_smiling_face:


sampettersson:
@FredrikAreschoug I am a bit concerned that we are making things a bit too explicit, I think we for example could split this notion and the discount notion into two metadatas, for example:

`MetadataBackgroundGradient` == .gradientOne
`MetadataPill` == "25% discount"

PR #396, opened 2021-01-21
Pill and background

FredrikAreschoug:
Added `EmbarkStoryMetadataEntryPill` and `EmbarkStoryMetadataEntryBackground` and deprecated `EmbarkStoryMetadataEntryDiscount`


styrelius:
These changes/additions are just for the app onboarding right? Since the intro screen for the web onboarding doesn't come from Embark. Or am I misunderstanding something? 🤔

FredrikAreschoug:
@styrelius yes for now it's only for the app but if you want to start using metadata to build this page up we can add more fields :slightly_smiling_face:

schema.ts
```diff
@@ -450,6 +456,21 @@ const typeDefs = `
       WEB_ONBOARDING
       APP_ONBOARDING
     }
+    
+    type EmbarkStoryMetadataEntryPill {
+      """Localized"""
+      pill: String!
```
sampettersson, line 462: Question, are these resolved server side? :)

sampettersson (dismissed):
Looks GTM! As long as we fetch and populate the text keys on the server :)

schema.ts
```diff
@@ -450,6 +456,21 @@ const typeDefs = `
...
+    type EmbarkStoryMetadataEntryPill {
+      """Localized"""
+      pill: String! 
+    }
+    
+    type EmbarkStoryMetadataEntryBackground {
+      background: EmbarkStoryMetadataEntryBackgroundEnum!
+    }
+    
+    enum EmbarkStoryMetadataEntryBackgroundEnum {
```
sampettersson, line 469: Maybe name this something like `...Option` instead of `...Enum`? as that becomes a bit ambiguous imo :)

schema.ts
FredrikAreschoug: yes they are :) that is what's happening on line 50-58 in `resolveStoriesMetadata.ts`

sampettersson (approved):
Dope! Thx 😄

PR #401, opened 2021-02-08
Allow `customData` to be null

hedvigoscar:
This aligns the schema with reality, as `customData` is not required to be specified.


sampettersson (approved):
HAHA I was just about to do this 😂

PR #404, opened 2021-02-23
Update to Web Onboarding DK - Contents (Add computed value for coInsured)

gustaveen:


styrelius (approved):
This looks like it should do the trick yes! 🙌 ⭐

PR #407, opened 2021-02-26
[GRW-12] Remove computedStoreValues as a prop on EmbarkProvider

gustaveen:
## What
Remove computedStoreValues as a prop on EmbarkProvider. This was the reason it didn't work in the web-onbarding where we didn't send it in as a prop in the `<EmbarkProvider>`.

## Why
To remove the risk of implementing this and forgetting to add the `computedStoreValues` prop (since it's derived from the data object anyway). It feels easier to encapsulate it it the actual component instead.

👍 or 👎

### DIFF
--- a/src/Components/EmbarkProvider.tsx
+++ b/src/Components/EmbarkProvider.tsx
@@ -21,6 +21,15 @@ interface EmbarkProviderProps {
   initialStore?: Store
 }
 
+const mapComputedStoreValuesList = (
+  computedStoreValues: ReadonlyArray<ComputedStoreValues> | undefined,
+) => {
+  return computedStoreValues?.reduce(
+    (acc, { key, value }) => ({ ...acc, [key]: value }),
+    {},
+  )
+}
+
 const StoreListener: React.FunctionComponent<EmbarkProviderProps> = (props) => {
   const { store } = React.useContext(StoreContext)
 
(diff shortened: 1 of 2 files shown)


src/Components/EmbarkProvider.tsx
```diff
@@ -34,19 +33,36 @@ const StoreListener: React.FunctionComponent<EmbarkProviderProps> = (props) => {
...
+    if (props.data.computedStoreValues) {
+      const parsedComputedStoreValues = (props.data
+        .computedStoreValues as ReadonlyArray<ComputedStoreValues>).reduce(
+        (acc, { key, value }) => ({ ...acc, [key]: value }),
+        {},
+      )
+      setComputedStoreValues(parsedComputedStoreValues)
+    }
+  }, [props.data.computedStoreValues])
+
```
palmenhq, line 51: just asking should this really be an effect? That means computed store values will be empty by default and may cause a flash with content right? could it be more suitable with some kind of memo case?

palmenhq (commented):
Good find! Just one thought really

src/Components/EmbarkProvider.tsx
```diff
@@ -19,7 +19,19 @@ interface EmbarkProviderProps {
...
   initialStore?: Store
-  computedStoreValues?: ComputedStoreValues
+}
+
+const mapComputedStoreValuesList = (
+  computedStoreValues: ReadonlyArray<ComputedStoreValues>,
+) => {
+  if (!computedStoreValues) {
+    return
+  }
```
palmenhq, line 29: isn't this redundant? Shouldn't `computedStoreValues` always be an array? Otherwise i think we should change the typing

gustaveen: I see what you mean. Since `data` is of type any `computedStoreValues` can be undefined right. And I don't see why typescript doesn't complain on that... But yeah, it should rather be this then:
```
const mapComputedStoreValuesList = (
  computedStoreValues: ReadonlyArray<ComputedStoreValues> | undefined,
) => {

  return computedStoreValues?.reduce(
    (acc, { key, value }) => ({ ...acc, [key]: value }),
    {},
  )
}
```

palmenhq: Yes that makes sense. Yes this suggestion looks good imo

palmenhq (approved):
😍

PR #409, opened 2021-03-08
Update to moving-flow-SE

zakkjellin:
First iteration of the Swedish moving flow! Yay!!!


angel-data/moving-flow-SE.json
```diff
@@ -0,0 +1,499 @@
+{
+  "id": "0e9214ae-d578-43b1-8e67-cdbb5aeddfd3",
+  "ifid": "29F1BC05-1702-4DA0-B866-9139E5B2C621",
+  "keywords": {
+    "isTrusted": true
+  },
```
gustaveen, line 6: @hedvigoscar @sampettersson is this a new keyword you added? or do you know where it comes from? :) keywords should only be strings right? and this one seems to pop up when ever I create a new story in angel 😅

hedvigoscar: No idea unfortunately 😞

gustaveen: The mystery continues! 🕵️‍♂️

styrelius: @gustaveen Won't this `isTrusted` make this flow break? So this flow needs to be remade or? 

Also - I think when it comes to the moving flow stuff it makes most sense that @sampettersson and @hedvigoscar does the approving right?

sampettersson: @gustaveen weird... went into Angel imported the file and opened the keywords modal and clicked save and it removed it?!

Screenshot 2021-03-25 at 19 23 15

@styrelius yeah, I agree, makes most sense if we review these 😄

angel-data/moving-flow-SE.json
```diff
@@ -0,0 +1,499 @@
+{
+  "id": "0e9214ae-d578-43b1-8e67-cdbb5aeddfd3",
+  "ifid": "29F1BC05-1702-4DA0-B866-9139E5B2C621",
+  "keywords": {
+    "isTrusted": true
```
sampettersson, line 5:
```suggestion
```

angel-data/moving-flow-SE.json
```diff
@@ -0,0 +1,498 @@
+{
+  "id": "0e9214ae-d578-43b1-8e67-cdbb5aeddfd3",
+  "ifid": "29F1BC05-1702-4DA0-B866-9139E5B2C621",
+  "keywords": {
```
sampettersson, line 4:
```suggestion
  "keywords": {},
```

angel-data/moving-flow-SE.json
```diff
@@ -0,0 +1,498 @@
...
+  },
```
sampettersson, line 5:
```suggestion
```

sampettersson (approved):

PR #415, opened 2021-03-09
GRW-235 Reverse birth date in Web Onboarding DK - Contents

gustaveen:
Birth date should be in reversed order for DK (dd-mm-yyyy)


styrelius (approved):
🎂

PR #421, opened 2021-03-22
Update to Web Onboarding DK Contents - Add text keys for story keywords

gustaveen:


angel-data/Web Onboarding DK - Contents.json
```diff
@@ -7,8 +7,12 @@
   ],
   "id": "892533f3-934e-40ac-bccf-238e15f634e2",
   "ifid": "2F54B9F2-B6C1-4CFC-BF17-4DB41DCB1ED0",
-  "keywords": {},
-  "lastUpdate": "2021-03-09T15:22:53.257Z",
+  "keywords": {
+    "backButton": "{DK_STORY_KEYWORDS_BACK_BUTTON}",
+    "selectActionSelectLabel": "{DK_STORY_KEYWORDS_SELECT_ACTION_LABEL}",
+    "tooltipModalInformationLabel": "{DK_STORY_KEYWORDS_TOOLTIP_MODAL_INFORMATION_LABEL}"
```
gustaveen, line 13: Or come to think of it, should we have these generic for all markets? Since the would be directly translated (like "Back", "Select" and such) 
```suggestion
    "backButton": "{STORY_KEYWORDS_BACK_BUTTON}",
    "selectActionSelectLabel": "{STORY_KEYWORDS_SELECT_ACTION_LABEL}",
    "tooltipModalInformationLabel": "{STORY_KEYWORDS_TOOLTIP_MODAL_INFORMATION_LABEL}"
```

styrelius (approved):
Nice! Really makes sense to have these text keys not market specific (since the keywords aren't and they should just be translated) - good thinking! 🧠 ✨

PR #423, opened 2021-03-22
Update to Web Onboarding DK - Add text keys for story keywords

gustaveen:


styrelius (approved):
Feels kind of strange that we have to add story keywords for every story now that we can translate them with Lokalise but I guess it would be a pretty tedious operation to change them to be global 🤷‍♀️

PR #425, opened 2021-03-22
Update to Web Onboarding DK Contents - fix isStudent passage

gustaveen:
`isYouth` is for NO quotes, `isStudent` is used for DK quotes


styrelius (approved):
👩‍🎓

PR #431, opened 2021-03-24
Feature/add markdown support to <Message>

gustaveen:
## What 
Added support for having actual links in `<Message>` in Embark by resolving markdown either from Lokalise or from Angel. Currently we have unclickable urls which is unarguably a bad user experience 🙃  

Screenshot 2021-03-24 at 14 38 22

Had to modify `replacePlaceholders` to not return an array of strings but couldn't see why we didn't return a string in the first place. So if you know why, holla at me 😄 

## Storybook and example from Angel preview

https://user-images.githubusercontent.com/6661511/112321680-8a752d80-8cb0-11eb-9945-227cae89fe3e.mov


src/Components/Common.test.ts
```diff
@@ -0,0 +1,16 @@
...
+  const resolvedText = replacePlaceholders(store, text)
+  expect(resolvedText).toBe('Your name is John Doe')
+})
+
+it('replace placeholders with number', () => {
+  const store: Store = { livingSpace: '42', firstName: 'John', lastName: 'Doe' }
+  const text = '{livingSpace} square meters'
+  const resolvedText = replacePlaceholders(store, text)
+  expect(resolvedText).toBe('42 square meters')
+})
```
styrelius, line 16: 🌟

src/Components/Common.tsx
```diff
@@ -17,30 +17,28 @@ export interface Replacements {
...
   if (!matches) {
-    return []
+    return ''
   }
 
-  return matches.map((placeholder, index) => {
+  const resolvedPlaceholders = matches.map((placeholder) => {
     if (!getPlaceholderKeyRegex().test(placeholder)) {
       return placeholder
     }
```
styrelius, line 32: I know you didn't write this now but is this `if` block needed? Don't we return the `placeholder` regardless on line 38?

styrelius (commented):
This is awesome Gustav!! 🔗 👑 ✨ 

I should maybe add that I of course have no idea why we would want to return an array of strings from `replacePlaceholders` 🤷‍♀️ But maybe @palmenhq knows?

src/Components/Common.test.ts
styrelius: To clarify ☝️ I just think you're a star for adding these tests!

src/Components/Common.tsx
gustaveen: It wasn't so obvious for me either but removing it brake things, so that's why I added a test as well :) Since we split the text in to an array of strings `[ 'Your name is ', '{firstName}', ' ', '{lastName}' ]` we can have an empty string (or a comma) in there and that should be early returned, otherwise it breaks when doing the regex match for the replacement key :)

src/Components/Common.test.ts
gustaveen: Thank u! 🙏  It makes it a bit easier to understand what's going on at least :)

src/Components/Common.tsx
```diff
@@ -17,30 +17,28 @@ export interface Replacements {
...
   replacements: Replacements,
   text: string,
-) => {
+): string => {
   const matches = text.split(getPlaceholderRegex()).filter((value) => value)
-
   if (!matches) {
-    return []
+    return ''
   }
```
palmenhq, line 27: Is this needed?  `matches.map` would just map an empty array right?

src/Components/Common.tsx
palmenhq, line 32: This if i believe is needed as we're making a non-null assertion on L36? `placeholder.match(getPlaceholderKeyRegex())` would return `null` if there's no match and `['the match']` if there is a match, causing the `[0]` to fail if there's no match. However maybe it'd be prettier to check the null value..

palmenhq (commented):
Awesome stuff! And awesome with the tests!

Just a general question though, what's your thoughts on just using `dangerouslySetInnerHtml` in this case? I feel like using `react-markdown` to render a link is a bit like using a bazooka for a toy gun problem. But maybe we would like to use other features of markdown that are clunkier to do with "just" html? Idk what's right, i'm just thinking react-markdown might have a quite an impact on bundle size. Or not, idk haven't checked but maybe we should consider it :) also as all updates to translations and content goes through code review right?

gustaveen:
> Awesome stuff! And awesome with the tests!
> 
> Just a general question though, what's your thoughts on just using `dangerouslySetInnerHtml` in this case? I feel like using `react-markdown` to render a link is a bit like using a bazooka for a toy gun problem. But maybe we would like to use other features of markdown that are clunkier to do with "just" html? Idk what's right, i'm just thinking react-markdown might have quite an impact on bundle size. Or not, idk haven't checked but maybe we should consider it :) also as all updates to translations and content goes through code review right?

Thanks! Good point about the bundle size! 👍 You're right on that point, we would probably not use so many other features of markdown as I see now. But I really see a benefit from writing links in markdown as we do in all other cases in Lokalise. Because using `dangerouslySetInnerHtml` would mean we would write html in Lokalise? Or did you just mean we could use another markdown parser like `marked` that's half the size but render it like `dangerouslySetInnerHtml`? :) @palmenhq

palmenhq:
@gustaveen true, i guess we could just do this on the server side in the parse phase right? :D

palmenhq:
Also FYI #432 fixes the failing build

gustaveen:
For sure, that would be nice to have in the parser! 🙌  I kinda don't want to derail the other DK tasks since this was not planned work so maybe we should pause this and continue on it later. Or do you have a super clear solution you want to add? @palmenhq 🙂

sampettersson:
We use markdown in other places in the apps already, so I think we could easily work with markdown on iOS and Android, do you see any risks @hedvigoscar @hugokallstrom?

hedvigoscar:
Not really, I think this is probably fine. The one concern I have is that when we introduce changes like these, it would be nice to be able to flag for it so that we can prompt users to upgrade their apps, so that they don't get unprocessed markdown rendered in their apps, for example.

Given that we intend to release the first iteration of embark to production in the coming week, that might be relevant sooner than later.

Implementation-wise though, it's very simple from an android-perspective 🙂

gustaveen:
Good input! 🙌  @sampettersson  @hedvigoscar I might have jumped the gun a bit here, should we maybe plan for us to implement a markdown resolver (or only links if we feel that's the only thing we need) in the clients when we have time and also handle the use case where members have to upgrade the app :)

src/Components/Common.tsx
gustaveen: Very true, that could be removed 👍 @palmenhq

sampettersson:
@gustaveen we haven't shipped anything on iOS yet so its easy to add there if we just say that markdown is something we wanna support :)

sampettersson:
@gustaveen FYI we are shipping iOS with markdown support 🎉

hedvigoscar:
android now supports Markdown. Hence, I'd like to merge this PR 🙂 @gustaveen

styrelius (dismissed):
🙃 We did say we should merge this, right?

hedvigoscar:
android concurs, we have markdown-support already 🙂

package.json
```diff
@@ -58,6 +58,7 @@
     "react-dom": "^16.13.0",
     "react-hot-loader": "^4.12.19",
     "react-input-mask": "^2.0.4",
+    "react-markdown": "^5.0.3",
```
robinandeer, line 61: haha I can't wait for Server Components to enable these features without hurting bundle size!

robinandeer (commented):
Does this help us with the Footer-component? Or is this a first step and then we can use Markdown in other places later?

src/Components/Common.tsx
```diff
@@ -80,6 +75,10 @@ export const MessageBody = styled.p<MessageBodyProps>`
   @media (max-width: 320px) {
     font-size: 14px;
```
robinandeer, line 76: nit:

```suggestion
    font-size: 0.875rem;
```

robinandeer (approved):
Code looks good! :D

PR #455, opened 2021-04-13
Add CODEOWNERS

sampettersson:
This would make it so that the GitHub team `clients` get auto picked as reviewers & someone from that team needs to approve it when a pr is created, in this way I hope information about embark features get widely distributed to all parties who might be affected 😄


hedvigoscar (dismissed):

styrelius (approved):
🧠

PR #457, opened 2021-04-20
Update to Web Onboarding - Swedish Needer

hugokallstrom:
Use `ExternalRedirect` with key `quoteId` to avoid having to hard code this in clients.

### DIFF
--- a/angel-data/Web Onboarding - Swedish Needer.json
+++ b/angel-data/Web Onboarding - Swedish Needer.json
@@ -29,7 +29,7 @@
     "selectActionSelectLabel": "Välj",
     "tooltipModalInformationLabel": "Information"
   },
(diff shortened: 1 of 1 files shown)


palmenhq (commented):
I think we don't have this for web currently, right @gustaveen @styrelius ? Do we need to implement it before merging this?

styrelius:
> I think we don't have this for web currently, right @gustaveen @styrelius ? Do we need to implement it before merging this?

@palmenhq  The way I understood it this what @sampettersson  fixed with [this PR](https://github.com/HedvigInsurance/web-onboarding/pull/498). And we already have this change in all Norwegian flows :)

sampettersson (approved):

sampettersson:
@palmenhq @styrelius oui! It works in web-onboarding too :)

PR #458, opened 2021-04-22
Deprecate returning `EmbarkMessage` from `EmbarkResponse`

hedvigoscar:
They are entirely equivalent. Might as well only have one


hugokallstrom (approved):
LGTM

sampettersson (approved):

PR #459, opened 2021-04-22
Update to Web Onboarding DK - Contents - add correct key for living space

gustaveen:
This makes sure the user don't complete the flow and hit the limits of UWGL


hugokallstrom (approved):
LGTM!

PR #460, opened 2021-04-22
Update to Web Onboarding DK - Danish Contents-Accident - add correct key for living space

gustaveen:
Also add correct translation key for select label


styrelius (approved):
👍 

Also I think there's a risk that there are more errors we're not catching correctly in the Danish flow(s), might be a good idea to go through that? Maybe 🤔

gustaveen:
very true @styrelius, I'm going over that now! 🙌  so it might be a PR bonanza :D

PR #464, opened 2021-04-23
Update to Web Onboarding - Swedish Needer - fix bug with quoteId being set to null

gustaveen:
This caused quoteIds to be set to null when going to offerpage


styrelius (approved):
🚒

PR #465, opened 2021-04-28
Make sure long placeholders don't make the input filed too wide

gustaveen:
When adding a longer placeholder for address field, the input calculation got too wide. Set a max size so it doesn't become too large, the content will still scroll if it's a super duper long address

**Without max size**
Screenshot 2021-04-28 at 15 15 23

**With max size**
Screenshot 2021-04-28 at 15 15 42


styrelius (approved):
👍 📏

PR #491, opened 2021-05-17
Add missing dependency to Dockerfile

hugokallstrom:
Docker was not able to build since the module `load-story.ts` was missing.


hedvigoscar (approved):
🚀

PR #502, opened 2021-05-20
EmbarkMultiActionNumberAction

hugokallstrom:
Add `EmbarkMultiActionNumberAction` which, compared to `EmbarkNumberAction` does not contain links, since those are present in the `MultiAction` parent.

### DIFF
--- a/src/Parsing/parseStoryData.ts
+++ b/src/Parsing/parseStoryData.ts
@@ -128,28 +128,11 @@ const getSelectAction = (
   }
 }
 
-const getMultiActionNumberAction = (
+const getNumberAction = (
   numberActionNode: Element,
   translate: Translator,
+  typeName: string,
 ) => {
-  const placeholder = translate(
-    numberActionNode.getAttribute('placeholder') || '',
-  )
-  const key = numberActionNode.getAttribute('key')
-  const unit = translate(numberActionNode.getAttribute('unit') || '')
-
-  return {
-    __typename: 'EmbarkMultiActionNumberAction',
-    component: 'MultiActionNumberAction',
-    data: {
-      placeholder,
-      key,
-      unit,
-    },
-  }
-}
-
-const getNumberAction = (numberActionNode: Element, translate: Translator) => {
   const placeholder = translate(
     numberActionNode.getAttribute('placeholder') || '',
   )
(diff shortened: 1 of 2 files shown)


src/Parsing/parseStoryData.ts
```diff
@@ -225,7 +246,7 @@ const getMultiAction = (multiActionNode: Element, translate: Translator) => {
```
hugokallstrom, line 246: Should we rename the node tag as well?

hedvigoscar: I think it's totally fine for the angel-level tag to be called `NumberAction`. This is more of an implementation-detail 🙂

src/Parsing/parseStoryData.ts
```diff
@@ -225,7 +246,7 @@ const getMultiAction = (multiActionNode: Element, translate: Translator) => {
 
   Array.from(addNode.getElementsByTagName('numberaction')).forEach(
     (numberActionNode) => {
-      components.push(getNumberAction(numberActionNode, translate))
```
hedvigoscar, line 228: If we want, we can re-use `getNumberAction`, and then switch out the `__typename`. In that way, we can reduce the code duplication. This will work as the `EmbarkMultiActionNumberAction` will likely always be a subset of `EmbarkNumberAction`. We can possibly express it as such:

```ts
components.push({...getNumberAction(numberActionNode, translate), __typename: 'EmbarkMultiActionNumberAction'})

File: src/Parsing/parseStoryData.ts [hunk 3]
```diff
@@ -128,6 +128,27 @@ const getSelectAction = (
...
+) => {
+  const placeholder = translate(
+    numberActionNode.getAttribute('placeholder') || '',
+  )
+  const key = numberActionNode.getAttribute('key')
+  const unit = translate(numberActionNode.getAttribute('unit') || '')
+
+  return {
+    __typename: 'EmbarkMultiActionNumberAction',
+    component: 'MultiActionNumberAction',
```
hedvigoscar, line 143: I think `component` generally just contains the source code for the component, weirdly enough 😬

hedvigoscar (approved):

src/Parsing/parseStoryData.ts
hugokallstrom: Cool, fixed!

hedvigoscar (approved):

PR #503, opened 2021-05-20
Vertically center Hedvig logo

robinandeer:
## What

Remove white space from bottom of svg.

## Why

The logo has some white space at the bottom so it doesn't look vertically centered.

https://user-images.githubusercontent.com/1220232/118952941-a87fa700-b95c-11eb-8f32-591fa7898936.mov

### DIFF
--- a/src/Components/Header.tsx
+++ b/src/Components/Header.tsx
@@ -34,6 +34,7 @@ const LogoImage = styled.img`
 
 const Svg = styled.svg`
   fill: currentColor;
+  vertical-align: bottom;
 `
 
 export const HedvigLogo: React.FunctionComponent = () => (


src/Components/Header.tsx
```diff
@@ -34,6 +34,7 @@ const LogoImage = styled.img`
 
 const Svg = styled.svg`
   fill: currentColor;
+  vertical-align: bottom;
```
gustaveen, line 37: Works perfect! Setting `display: flex` on the parent `<a>` could be an alternative

gustaveen (approved):
So nice that you can see stuff like that, I'm blind to that by now haha :D

src/Components/Header.tsx
robinandeer: Nice, I took inspiration from here: https://stackoverflow.com/questions/7774814/remove-white-space-below-image

I thought this was the least intrusive way

gustaveen: Ah nice, so it actually aligns it along the baseline, I think it's just the semantics of `bottom` that confuses me :D

PR #504, opened 2021-05-20
Add label parameter to Multi Action Number Action

tarikstafford:


schema.ts
```diff
@@ -266,6 +266,7 @@ const typeDefs = `
         key: String!
         placeholder: String!
         unit: String
+        label: String!
```
hugokallstrom, line 269: Should this be non-nullable? This is the current structure in angel: 
```
 <NumberAction placeholder="20"
                  key="area"
                  unit="{EMBARK_MOVING_SE_APARTMENT_SQUARE_METERS}"
    >
    </NumberAction>
```

tarikstafford: Screenshot 2021-05-20 at 14 22 53

Would be weird without a title right?

tarikstafford: You are right though if a string isn't found it should be nullable.

hugokallstrom (approved):

PR #513, opened 2021-06-04
Trim DropdownAction Text

sampettersson:
Because the value of text was something like `\n        Other\n      `


hugokallstrom (approved):
Should we fix the text at the source as well?

tarikstafford:
> Should we fix the text at the source as well?

Think that makes sense, but maybe need to confer with web first.

sampettersson:
@hugokallstrom @tarikstafford That is expected, this is how angel formats itself, so the dom node would contain the `\n`, this is the approach we used before.

PR #514, opened 2021-06-16
Feature/address autocomplete

robinandeer:
Add address autocomplete action component for Danish addresses.

## Additional updates
- Add Apollo Client to test with real API during development.
- Integrate mock APIs with Storybook
- Update babel setup so we can use this syntax:
  
  ```
  variable ?? 'defaultValue'
  ```
- Add Modal component that works better for this use case compared to the global Modal component.
- Implement combobox/autocomplete with Downshift lib.

## Preview

https://user-images.githubusercontent.com/1220232/122754736-4fc16800-d294-11eb-9c59-d1b789db0819.mov


src/Components/API/addressAutocomplete.ts
```diff
@@ -0,0 +1,34 @@
...
+  address: string
+  streetName?: string
+  streetNumber?: string
+  floor?: string
+  apartment?: string
+  postalCode?: string
+  city?: string
+}
+
+export type CompleteAddress = AddressSuggestion & {
```
gustaveen, line 12: This might be the way to do it, just feels a bit weird. Since the intersection type would be right?:

```
type CompleteAddress = {
  id?: string
  address: string
  streetName?: string
  streetNumber?: string
  floor?: string
  apartment?: string
  postalCode?: string
  city?: string
  id: string
  address: string
  streetName: string
  streetNumber: string
  postalCode: string
  city: string
}
```

Or does it overwrite existing, optional parameter like `streetName?:` to being required `streetName:`.

src/Components/Actions/AddressAutocompleteAction/Modal.tsx
```diff
@@ -0,0 +1,106 @@
+import * as React from 'react'
+import { motion } from 'framer-motion'
+import { colorsV3 } from '@hedviginsurance/brand'
+import styled from '@emotion/styled'
+
+const StyledOverlay = styled(motion.div)`
```
gustaveen, line 6: A general thought about using `Styled` as prefix. Personally I think we can omit it and just have `Overlay`. I think it's easy enough to just inspect or click the component to see if if it's a styled component or not.

src/Components/Actions/AddressAutocompleteAction/Modal.tsx
```diff
@@ -0,0 +1,106 @@
...
+  React.useEffect(() => {
+    // Prevent background from scrolling behind modal
+    if (isOpen) {
+      document.body.style.overflow = 'hidden'
+    }
+
+    return () => {
+      document.body.style.overflow = 'initial'
+    }
+  }, [isOpen])
```
gustaveen, line 83: Nice and simple! 👌

src/Components/Actions/AddressAutocompleteAction/Modal.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+    return () => {
+      document.removeEventListener('mousedown', handleClickOutside)
+      document.removeEventListener(`touchstart`, handleClickOutside)
+    }
+  }, [ref, onDismiss])
+
+  React.useEffect(() => {
+    const handleEscape = (event: KeyboardEvent) => {
+      if (event.key === 'Escape') {
```
gustaveen, line 63: Awesome that you added esc as well! 🙌

src/Components/Actions/AddressAutocompleteAction/useDebounce.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    return () => {
+      clearTimeout(handler)
+    }
+  }, [value, delay])
+
+  return debouncedValue
+}
+
+export default useDebounce
```
gustaveen, line 25: Neat! 🙏

gustaveen:
Amazing work on this @robinandeer! 🏗 🎉

src/Components/Actions/AddressAutocompleteAction/Modal.tsx
styrelius: I agree 👍

src/Components/API/addressAutocomplete.ts
robinandeer: Yes, that is what I assumed and TS seems to pick it up that way. However, if there's another approach - I would be happy to change it!

src/Components/Actions/AddressAutocompleteAction/Modal.tsx
robinandeer: Yeah, I think it was a convention we used in a different project. But I've removed the prefix now!

src/Components/Actions/AddressAutocompleteAction/Modal.tsx
robinandeer: it comes automatically as soon as I deal with modals - I only know how frustrated I would be without it working haha

src/Components/Actions/AddressAutocompleteAction/useDebounce.ts
robinandeer: Thanks! I think it's much more intuitive to debounce a value rather than a function somehow. That's why I wrote this instead of using lodash debounce.

robinandeer:
Thanks @gustaveen! :D Still think we can do quite a lot to make it extra readable/maintainable - so highlight anything strange you come across!

@sampettersson has also promised to "roast" our code so we learn to develop Embark the "cannonical" way ;)

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -0,0 +1,580 @@
...
+  text-align: left;
+  width: 100%;
+  position: relative;
+
+  padding: 8px 16px;
+  border-radius: 8px;
+  border: 1px solid
+    ${(props) => (props.focus ? colorsV3.purple300 : colorsV3.gray500)};
+  ${(props) => props.focus && `box-shadow: 0 0 0 2px ${colorsV3.purple300}`}
+`
```
gustaveen, line 130: This works perfectly fine, just sharing an alternative solution for focus state that's kind of neat IMO :) With `:focus-within` you can omit the `isInputFocus` state and write it like so. Not supported in IE though 

```suggestion
const ComboboxField = styled.div`
  box-sizing: border-box;
  text-align: left;
  width: 100%;
  position: relative;

  padding: 8px 16px;
  border-radius: 8px;
  border: 1px solid colorsV3.gray500;

   &:focus-within {
    border-color: ${colorsV3.purple500};
    box-shadow: 0 0 0 2px ${colorsV3.purple300}
   }  
`
```

I would assume Mange and Zak would like to have a darker tint of gray instead of purple for the focus state, but we can go over that later 👍

robinandeer: Nice! I like that solution -- and at this point, non-critical CSS should not have to consider IE

package.json
```diff
@@ -83,7 +84,9 @@
     "react-dom": "16.13.x"
   },
   "devDependencies": {
+    "@apollo/client": "^3.3.20",
     "@babel/core": "^7.10.4",
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.14.5",
```
robinandeer, line 89: So we can use the `variable ?? 'defaultValue'` syntax

schema.ts
```diff
@@ -5,7 +5,7 @@ import { promises } from 'fs'
 import { loadStory } from './load-story'
 import { resolveMetadataOnLocale } from './src/Resolvers/resolveStoriesMetadata'
 
-const typeDefs = `
+const typeDefs = /* GraphQL */ `
```
robinandeer, line 8: Get syntax highlighting in VSCode

src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
```diff
@@ -0,0 +1,238 @@
...
+  selected: AddressSuggestion | null
+  onSelect: (suggestion: AddressSuggestion | null) => void
+  onNotFound: () => void
+  placeholder: string
+  value: string
+  onChange: (newValue: string) => void
+  onClear: () => void
+}
+
+const AddressAutocomplete: React.FC<Props> = (props) => {
```
robinandeer, line 99: This component is supposed to handle address autocomplete. It's not aware of anything Embark-related (other than getting translations).

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -0,0 +1,305 @@
...
+  }
+
+  if (address.apartment) {
+    data[STORE_KEY.APARTMENT] = address.apartment
+  }
+
+  return data
+}
+
+export const AddressAutocompleteAction: React.FC<AddressAutocompleteActionProps> = (
```
robinandeer, line 136: This is the Embark-layer. It is aware of all the Embark-logic and integrates the address autocomplete component.

src/Components/API/ApiContext.tsx
```diff
@@ -81,6 +86,33 @@ export const mockApiResolvers: TApiContext = {
     await timeout(300)
     return houseInformationMocks[0]
   },
+  addressAutocompleteQuery: async (term) => {
+    const { ApolloClient, InMemoryCache, gql } = require('@apollo/client')
+
+    const client = new ApolloClient({
+      uri: 'https://graphql.dev.hedvigit.com/graphql',
```
sampettersson, line 93: Maybe move this to `.env`? I think we have this in at least two places now.

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -0,0 +1,305 @@
...
+  formatAddressLine,
+  formatAddressLines,
+  isCompleteAddress,
+  isSameAddress,
+} from './utils'
+import AddressAutocomplete from './AddressAutocomplete'
+
+const ADDRESS_NOT_FOUND = 'ADDRESS_NOT_FOUND'
+
+const STORE_KEY = {
```
sampettersson, line 25: Make sure to document these in `embark-docs` 🤠

sampettersson (commented):
👯

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
robinandeer: We have a PR ongoing for that! Perhaps we should've name spaced it as well to avoid collisions

src/Components/API/ApiContext.tsx
robinandeer: @sampettersson We don't have dotenv integrated yet... is it worth adding or should we just centralize these in a config module?

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -187,10 +187,20 @@ export const AddressAutocompleteAction: React.FC<AddressAutocompleteActionProps>
...
+        // Reset confirmed address unless it matches the updated search term
+        if (prevValue && newValue !== formatAddressLine(prevValue)) {
+          return null
+        }
+
+        return prevValue
+      })
+    },
+    [store],
+  )
```
gustaveen, line 203: 👌

src/Components/Actions/AddressAutocompleteAction/useAddressSearch.ts
```diff
@@ -0,0 +1,62 @@
...
+  AddressSuggestion[] | null,
+  React.Dispatch<React.SetStateAction<AddressSuggestion[] | null>>,
+] => {
+  const api = React.useContext(ApiContext)
+  const [suggestions, setSuggestions] = React.useState<
+    AddressSuggestion[] | null
+  >(null)
+
+  const debouncedSearchTerm = useDebounce(searchTerm, 300)
+  const apiQuery = React.useMemo(
```
styrelius, line 36: What's the reason for using `useMemo` here? Won't the value returned from `getApiQuery` change all the time? Or am I missing something? 😅

src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
```diff
@@ -0,0 +1,236 @@
...
+    getComboboxProps,
+    highlightedIndex,
+    getItemProps,
+  } = useCombobox<AddressSuggestion>({
+    selectedItem: selected,
+    inputValue: value,
+    items: comboboxItems,
+    itemToString: (item) => (item ? formatAddressLine(item) : ''),
+    onInputValueChange: ({ inputValue, type }) => {
+      if (type === '__input_keydown_escape__') return
```
gustaveen, line 146: Ah nice, so you get these change types from combobox 👌

gustaveen (approved):
Boom! 5+ work Robin! ⭐️  Tested it out locally as well, works really good so let's get this out and test and iterate 🙌

src/Components/Actions/AddressAutocompleteAction/useDebounce.ts
```diff
@@ -0,0 +1,25 @@
+import { useState, useEffect } from 'react'
+
+/**
+ * React hook to debounce a value for a given interval
+ * @param value Value to debounce
+ * @param delay Delay interval in milliseconds
+ * @returns The same value
+ */
+const useDebounce = <T = any>(value: T, delay: number): T => {
```
styrelius, line 9: Why does the `T` have to have `any` as default type? 

Also, maybe `T` could rather be called something like `ValueType`? I think it's nice when you're a bit more descriptive in generics 🙃

styrelius (commented):
Amazing work @robinandeer !! 👏 ✨ 🤩

src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
robinandeer: Don't really understand why it's called on escape though -- the value isn't changing 😅

src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
```diff
@@ -0,0 +1,238 @@
+import * as React from 'react'
```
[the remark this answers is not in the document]
robinandeer: I was just following the conventions in the lib -- but seems like we can actually use the simple import statement as well. It's to do with the TS setup:

https://www.reddit.com/r/typescript/comments/dy0ix8/import_as_react_vs_import_react/

src/Components/Actions/AddressAutocompleteAction/Combobox.tsx
```diff
@@ -0,0 +1,128 @@
+import styled from '@emotion/styled'
+import { colorsV3, fonts } from '@hedviginsurance/brand'
```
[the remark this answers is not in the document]
robinandeer: Yeah, I wonder how that works when we package embark 🤔 

It's taken care of by web-onboarding, no? But we should try to keep them in sync still!

src/Components/Actions/AddressAutocompleteAction/useAddressSearch.ts
robinandeer: Not if you stop typing 😉 

Whenever I have a "use effect" dependency I want to memorize it to be sure. But in this case we have a primitive type (string) so we don't need it for that.

So just a minor optimization.

src/Components/Actions/AddressAutocompleteAction/useDebounce.ts
robinandeer: Yes, I remember we talked about this - I agree with you!

Regarding the default type, we could set it to "unknown" perhaps or we just make it mandatory?

src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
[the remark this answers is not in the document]
gustaveen: No worries, embark haven't got the all the love as the other repos 😁

src/Components/Actions/AddressAutocompleteAction/Combobox.tsx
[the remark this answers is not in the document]
styrelius: Ah, yes, maybe Embark just uses that from the `web-onboarding` 🧠 Unfortunately I have pretty large knowledge gaps when it comes to peer dependencies and how packages actually work 🙈

src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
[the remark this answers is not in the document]
styrelius: Maybe just leave it then and do a more thorough cleaning of these kinds of things in Embark at some point? Or will we just never fix it then? 😄

src/Components/Actions/AddressAutocompleteAction/useDebounce.ts
styrelius: Seems to me like it works fine to just remove the default type? Or I tried it and nothing seemed to immediately break at least 😅

src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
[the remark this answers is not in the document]
robinandeer: Merge it with web-onboarding 😉

src/Components/Actions/AddressAutocompleteAction/useDebounce.ts
robinandeer: Sounds about right - I might've been to quick at implementing a stack overflow example :P

PR #528, opened 2021-06-24
Set a larger touch target for autocomplete items

gustaveen:
## What 
Larger autocomplete list items 

## Why
It looked good in Chrome but not in iOS Safari since we don't have a global `box-sizing` rule in embark 😄

### DIFF
--- a/src/Components/Actions/AddressAutocompleteAction/Combobox.tsx
+++ b/src/Components/Actions/AddressAutocompleteAction/Combobox.tsx
@@ -77,9 +77,10 @@ const ComboboxInput = styled.input`
 
 const ComboboxOption = styled.li`
   padding: 8px 16px;
-  min-height: 32px;
+  min-height: 48px;
   display: flex;
   align-items: center;
+  box-sizing: border-box;
 
   text-align: left;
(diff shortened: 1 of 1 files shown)


src/Components/Actions/AddressAutocompleteAction/Combobox.tsx
```diff
@@ -77,9 +77,10 @@ const ComboboxInput = styled.input`
 
 const ComboboxOption = styled.li`
   padding: 8px 16px;
-  min-height: 32px;
+  min-height: 48px;
   display: flex;
   align-items: center;
+  box-sizing: border-box;
```
robinandeer, line 83: Oh! That's the reason it looked strange. I've become so used that it's part of a CSS reset :P

I wonder if Storybook injects its own reset because I was confused that it looked different when deployed in Embark 🤔

robinandeer (approved):
Good fix!

src/Components/Actions/AddressAutocompleteAction/Combobox.tsx
gustaveen: Ah, that might be it! I got confused by that as well 😄

PR #529, opened 2021-06-28
Fix formatting of address line

robinandeer:
Update API query to limit search on apartment + floor.
Street name is sometimes abbreviated so need to use address field.

https://user-images.githubusercontent.com/1220232/123637479-5c0e6d80-d81e-11eb-8f1e-e7b2d32a9e88.mov


robinandeer:
Close in favor of #531

PR #530, opened 2021-06-29
Avoid cursor jumping in input field

robinandeer:
## What?

Move onChange handler to actual input element.
Update input value when selecting suggestion.
Call either onNotFound OR onSelect.
Call reset function on clear.
Respond to empty input field in parent component.
Skip passing `itemToString` and `selectedItem` from `useCombobox` (not needed).

## Why?

There's a known issue in Downshift relating to this: https://github.com/downshift-js/downshift/issues/1108. It happens when you are using a controlled input field. The reason is that the value is set asynchronously which is too late for React to handle the cursor correctly.

The suggested solution is to update the input value using the native onChange handler directly on the input element. This also means we need to update the input value when selecting an option. However, that almost feels more natural rather then letting Downshift handle that for us.

https://user-images.githubusercontent.com/1220232/123746428-ee138600-d8b1-11eb-8816-2f585088e083.mov

### DIFF
--- a/src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
+++ b/src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
@@ -132,35 +132,26 @@ const AddressAutocomplete: React.FC<Props> = (props) => {
     getComboboxProps,
     highlightedIndex,
     getItemProps,
+    reset,
   } = useCombobox<AddressSuggestion>({
-    selectedItem: selected,
     inputValue: value,
     items: comboboxItems,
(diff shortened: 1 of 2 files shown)


gustaveen (dismissed):
Aha, tricky one! But I agree, this feels easier in a way. Good job! 🙌

styrelius (approved):
Great job! 🎭  🏹  
Spotted an unused import in `AddressAutoComplete.tsx` (`formatAddressLine` on line 10) but since we don't have ESLint in embark what isn't caught. We should definitely do something about that at some point

gustaveen (approved):

PR #531, opened 2021-06-29
Integrate address autocomplete type in API

robinandeer:
## What

Pass adress autocomplete type based on picked suggestion.
Skip passing trailing space after street name.
Tweak api query string when searching for apartments.
Rename variables: "address" to "suggestion".

## Why

Be more explicit about what type of suggestions we are searching for: streets, buildings, or apartments.

## Demo

https://user-images.githubusercontent.com/1220232/123770739-cd0b5f00-d8ca-11eb-906b-a3995f2f73a3.mov

### DIFF
--- a/src/Components/API/addressAutocomplete.ts
+++ b/src/Components/API/addressAutocomplete.ts
@@ -17,3 +17,10 @@ export type CompleteAddress = AddressSuggestion & {
   postalCode: string
   city: string
 }
+
+export type AddressAutocompleteType = 'STREET' | 'BUILDING' | 'APARTMENT'
+
+export type AddressAutocompleteQuery = (
+  searchTerm: string,
+  options?: { type: AddressAutocompleteType },
+) => Promise<AddressSuggestion[]>
--- a/src/Components/Actions/AddressAutocompleteAction/useAddressSearch.ts
+++ b/src/Components/Actions/AddressAutocompleteAction/useAddressSearch.ts
@@ -46,9 +46,9 @@ const useAddressSearch = (
   >(null)
 
   const debouncedSearchTerm = useDebounce(searchTerm, 300)
-  const [apiQuery, suggestionType] = React.useMemo(
-    () => getApiQuery(debouncedSearchTerm, suggestion),
-    [debouncedSearchTerm, suggestion],
+  const [apiQuery, suggestionType] = getApiQuery(
+    debouncedSearchTerm,
+    suggestion,
   )
 
   // @ts-ignore: clean-up function only needed conditionally
--- a/src/Components/Actions/AddressAutocompleteAction/useAddressSearch.ts
+++ b/src/Components/Actions/AddressAutocompleteAction/useAddressSearch.ts
@@ -1,23 +1,36 @@
-import { AddressSuggestion } from '../../API/addressAutocomplete'
+import {
+  AddressSuggestion,
+  AddressAutocompleteType,
+} from '../../API/addressAutocomplete'
 import * as React from 'react'
 import { ApiContext } from '../../API/ApiContext'
 import { isMatchingStreetName } from './utils'
 import useDebounce from './useDebounce'
 
-const getApiQuery = (searchTerm: string, suggestion?: AddressSuggestion) => {
+const getApiQuery = (
+  searchTerm: string,
+  suggestion?: AddressSuggestion,
+): [string, AddressAutocompleteType] => {
   if (suggestion) {
-    if (suggestion.city === undefined) {
+    if (!suggestion.city) {
       // Refine search after selecting a suggested street name
-      return searchTerm + ' '
+      return [searchTerm, 'BUILDING']
     }
 
-    if (suggestion.postalCode && isMatchingStreetName(searchTerm, suggestion)) {
+    if (
+      suggestion.city &&
+      suggestion.postalCode &&
+      isMatchingStreetName(searchTerm, suggestion)
+    ) {
       // Refine search after selecting a specific building (floor & apartment)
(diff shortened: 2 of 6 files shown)


src/Components/Actions/AddressAutocompleteAction/useAddressSearch.ts
```diff
@@ -1,23 +1,36 @@
...
-    if (suggestion.postalCode && isMatchingStreetName(searchTerm, suggestion)) {
+    if (
+      suggestion.city &&
+      suggestion.postalCode &&
+      isMatchingStreetName(searchTerm, suggestion)
+    ) {
       // Refine search after selecting a specific building (floor & apartment)
-      return `${searchTerm} ${suggestion.postalCode} ${suggestion.city}`
+      return [
+        `${searchTerm}, , ${suggestion.postalCode} ${suggestion.city}`,
```
robinandeer, line 27: We still need to tweak the API query string a little when searching for apartments. That's why we are adding the extra commas here.

I kinds still prefer that to introduce the "caretpos" parameter, however 😛

src/Components/Actions/AddressAutocompleteAction/utils.ts
```diff
@@ -3,19 +3,19 @@ import {
...
-export const formatAddressLine = (address: AddressSuggestion): string => {
-  if (address.streetName && address.streetNumber) {
-    let displayAddress = `${address.streetName} ${address.streetNumber}`
-    if (address.floor) {
-      displayAddress += `, ${address.floor}.`
+export const formatAddressLine = (suggestion: AddressSuggestion): string => {
+  if (suggestion.streetName && suggestion.streetNumber) {
+    let displayAddress = `${suggestion.streetName} ${suggestion.streetNumber}`
+    if (suggestion.floor) {
+      displayAddress += `, ${suggestion.floor}.`
```
robinandeer, line 10: This is very DK-specific - I guess we are okey with it for now but I'm a little unhappy about how it looks now :P

gustaveen (dismissed):
Boom, that works perfect for all use cases I tested! Well done! 🙌 I also prefer this solution to adding `caretpos`

gustaveen (approved):

src/Components/Actions/AddressAutocompleteAction/useAddressSearch.ts
```diff
@@ -33,7 +46,7 @@ const useAddressSearch = (
   >(null)
 
   const debouncedSearchTerm = useDebounce(searchTerm, 300)
-  const apiQuery = React.useMemo(
+  const [apiQuery, suggestionType] = React.useMemo(
```
styrelius, line 49: Please remind me again - what was the purpose of memoizing this instead of just calling `getApiQuery` from inside of the `useEffect` that needs the return values and add `debouncedSearchTerm, suggestion` to the dependency array of that `useEffect`? 🙃

src/Components/Actions/AddressAutocompleteAction/utils.ts
```diff
@@ -3,19 +3,19 @@ import {
   CompleteAddress,
 } from '../../API/addressAutocomplete'
 
-export const formatAddressLine = (address: AddressSuggestion): string => {
-  if (address.streetName && address.streetNumber) {
-    let displayAddress = `${address.streetName} ${address.streetNumber}`
-    if (address.floor) {
-      displayAddress += `, ${address.floor}.`
+export const formatAddressLine = (suggestion: AddressSuggestion): string => {
```
styrelius, line 6: Maybe destructure the properties? 
```typescript
export const formatAddressLine = ({ 
  address, 
  streetName, 
  streetNumber, 
  floor, 
  apartment 
}: AddressSuggestion): string =>
```
Get's a little noisy to me with all the `suggestion.` everywhere 😅

styrelius:
And also, of course, GREAT job with all of this autocomplete stuff 👑 🌴 🏄 🍹

src/Components/API/addressAutocomplete.ts
```diff
@@ -17,3 +17,10 @@ export type CompleteAddress = AddressSuggestion & {
   postalCode: string
   city: string
 }
+
+export type AddressAutocompleteType = 'STREET' | 'BUILDING' | 'APARTMENT'
```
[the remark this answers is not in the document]
robinandeer: Yeah, not in Embark as far as I know - another reason to move the Embark web implementation to web-onboarding!

src/Components/Actions/AddressAutocompleteAction/useAddressSearch.ts
robinandeer: That's probably a force of habit on my part - it should not make a difference really since we are dealing with primitive types (string)

gustaveen:
Looks good to me! 🙌 I let @styrelius to review the changes 👍

src/Components/Actions/AddressAutocompleteAction/useAddressSearch.ts
styrelius: OK, so then the `useMemo` will rerun on every render instead? I still don't really understand, but never mind 😄

styrelius: And now I saw that you removed it so even more never mind haha 😅

styrelius (approved):
🤸‍♀️ 🤸‍♀️ 🤸‍♀️ 🍰

src/Components/Actions/AddressAutocompleteAction/useAddressSearch.ts
robinandeer: Yes, you are right that the `useMemo` would've most likely rerun on every render!

PR #532, opened 2021-06-29
Prevent input field from overflowing

robinandeer:
## What

Make sure the input field respects the max width and padding of it's containers.
Truncate text (ellipsis) for very long addresses.

## Why

This is what happens today:

Screenshot 2021-06-29 at 11 18 09

## Demo

Mobile

Screenshot 2021-06-29 at 11 21 53

Desktop + truncate text

Screenshot 2021-06-29 at 11 25 49

☝️ This is the one I'm not 100% sure on. Perhaps it's okey though in this case since it's more of a preview of your address and you can click it to view the full string.

### DIFF
--- a/src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
+++ b/src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
@@ -36,6 +36,7 @@ const STORE_KEY = {
 }
 
 const Container = styled.div`
+  max-width: 100%;
   display: flex;
   flex-direction: column;
   align-items: center;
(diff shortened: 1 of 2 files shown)


styrelius (commented):
I think truncating with ellipsis is fine but it feels a bit strange to me to do it on desktop where we actually have a lot of space horizontally 🤔  Is there a possibility to just let the "input" field grow even more if the address is super long? On desktop I mean, we should definitely use the  `text-overflow: ellipsis;` for mobile.

robinandeer:
It only truncates text if the input grows bigger than 100% of the screen width.

So the example is the odd use case of embark on Ipad in portrait mode basically.

On desktop this isn't much of a concern as you say. But I don't think it hurts to keep the ellipsis CSS rules for all screen sizes, no?

styrelius (approved):
Very good! ✂️ ✨

PR #533, opened 2021-06-29
Refactor AddressAutocomplete component

robinandeer:
- Refactor `AddressAutocomplete` to make it easier to test and maintain.
  - Move `Modal` + `useAddressSearch` hook to the "embark" integration layer.
  - Pass suggestions to the address autocomplete component.
  - Remove selected prop from component
  - Reset suggestions on select in parent component.

- Add tests for `AddressAutocomplete`

- Implement improvements to `Modal` so that we unmount it when it's not active
  - Simplify animation setup with AnimatePresence.
  - Remove isActive prop from AddressAutocomplete component.
  - Use `autoFocus` prop on input field.

No visual changes!

This is my view of the responsibilities of the different components/layers:

image


src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+        onNotFound={jest.fn()}
+        value=""
+        onChange={handleChange}
+        onClear={jest.fn()}
+        placeholder={placeholder}
+        suggestions={null}
+      />,
+    )
+
+    // Input is automatically focused
```
robinandeer, line 31: I wasn't sure how much to comment on what is asserted here.

How do you feel? Is it understandable already or do we need more comments?

src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+        onClear={handleClear}
+        placeholder=""
+        suggestions={null}
+      />,
+    )
+
+    act(() => {
+      userEvent.click(
+        getByRole('combobox').querySelector('button') as HTMLButtonElement,
+      )
```
robinandeer, line 108: I realized that it wasn't super clear how to target the button "as a user". Perhaps we are missing some a11y labels but works now.

src/Components/Actions/AddressAutocompleteAction/Modal.test.tsx
```diff
@@ -22,9 +22,9 @@ describe('address autocomplete modal', () => {
       )
     }
 
-    const { getByText } = render(<ModalWithButton />)
+    const { getByText, queryByText } = render(<ModalWithButton />)
 
-    expect(getByText('close')).not.toBeVisible()
+    expect(queryByText('close')).not.toBeInTheDocument()
```
robinandeer, line 27: The Modal contents are unmounted when the modal is not active now.

src/Components/Actions/AddressAutocompleteAction/Modal.tsx
```diff
@@ -84,35 +83,28 @@ const Modal: React.FC<Props> = ({ children, isOpen, onDismiss }) => {
...
-      >
-        {children}
-      </StyledModal>
-    </StyledOverlay>
+    <AnimatePresence>
+      {isOpen ? (
+        <StyledOverlay
+          initial={{ opacity: 0 }}
+          animate={{ opacity: 1 }}
+          exit={{ opacity: 0 }}
```
robinandeer, line 91: So much cleaner 💅

src/Components/KeywordsContext.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import * as React from 'react'
```
robinandeer, line 1: Needed for ts-jest until we include "esModuleInterop" setting in tsconfig. Should we add that already?

robinandeer:
New Modal animation:


https://user-images.githubusercontent.com/1220232/124738827-990eea00-df19-11eb-9db2-09a65c65478b.mov

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -237,47 +193,24 @@ export const AddressAutocompleteAction: React.FC<AddressAutocompleteActionProps>
...
-    clearStoreValues()
-
-    setValue(STORE_KEY.ADDRESS_SEARCH_TERM, searchTerm)
-    setValue(props.storeKey, ADDRESS_NOT_FOUND)
+    saveNotFound(searchTerm)
     props.onContinue()
-  }, [clearStoreValues, setValue, props.storeKey, searchTerm])
+  }, [searchTerm, saveNotFound])
+
+  const handleDismissModal = React.useCallback(
```
styrelius, line 210: Is there a specific reason for using `useCallback` here?

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -331,23 +264,32 @@ export const AddressAutocompleteAction: React.FC<AddressAutocompleteActionProps>
...
+          disabled={false}
+          text={(props.link || {}).label || 'Continue'}
+        />
+      ) : (
+        <ContinueButton
+          onClick={() => {}}
+          disabled={true}
+          text={(props.link || {}).label || 'Continue'}
+        />
+      )}
```
styrelius, line 279: Maybe just make the button `disabled` if there's no `confirmedAddress` instead of conditionally rendering buttons with different attributes?
```suggestion
        <ContinueButton
          onClick={() => handleContinue(confirmedAddress)}
          disabled={Boolean(confirmedAddress)}
          text={(props.link || {}).label || 'Continue'}
        />
```

styrelius: Or just with implicit type coercion if you like that better 🙃 
```typescript
        <ContinueButton
          onClick={() => handleContinue(confirmedAddress)}
          disabled={confirmedAddress}
          text={(props.link || {}).label || 'Continue'}
        />
```

src/Components/KeywordsContext.tsx
styrelius: Sure :) If that doesn't seem to mess up anything else.

src/Components/Actions/AddressAutocompleteAction/useStoreAddress.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  const initialAddress = React.useMemo(() => getAddressFromStore(store), [])
+
+  const clearStoreValues = React.useCallback(
+    () =>
+      Object.values(STORE_KEY).forEach((storeKey) => {
+        removeValues(storeKey)
+      }),
+    [removeValues],
+  )
```
styrelius, line 78: Would it be the same effect if this function was placed outside of the `useStoreAddress` hook (taking `removeValues` as an argument) as now when it's wrapped in a `useCallback`? If yes, I think the former is a nicer approach since it adds less complexity.

src/Components/KeywordsContext.tsx
robinandeer: A good task for Wednesday package update I think! So I'll keep it like this for now.

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
robinandeer: Just the very slight performance benefit of not regenerating it on every re-render. But could skip it.

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
robinandeer: The problem is that `confirmedAddress` is possibly null and therefore can't be passed directly to `handleContinue`.

We could use the "bang operator" to override this type check but I kind of prefer this explicit check instead. Where do you stand on that?

src/Components/Actions/AddressAutocompleteAction/useStoreAddress.ts
robinandeer: I'm not quite sure what approach you are aiming for 🤔 

`removeValues` comes from the "store" context. Do you mean it would be passed into `useStoreAddress` as a parameter? In that case both this hook and the parent component would have to subscribe to the "store" context. I tried to keep that logic contained within this hook.

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
styrelius: Ah, I see, I missed that! Hm, I think non-null assertions should probably be used sparingly (so no,  don't have an exciting opinion there). But how about this?
```typescript
        <ContinueButton
          onClick={confirmedAddress ? () => handleContinue(confirmedAddress) : () => {}}
          disabled={confirmedAddress}
          text={(props.link || {}).label || 'Continue'}
        />
```
Or is that also messy? I just think it's a bit confusing with the "two" buttons.

src/Components/Actions/AddressAutocompleteAction/useStoreAddress.ts
styrelius: I don't know - I just feel that all the `useCallback`s raises a lot of questions and make things complex. So I was just thinking if there could be some other approach - like moving some functions out from the hook.

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
robinandeer: Right, I agree that this is more what you would expect. I can change to that.

Interesting to figure out a good way to handle this though 🤔 like if the function should accept `null` as well but throw a runtime exception if it ever get's called with a "null" value maybe?

src/Components/Actions/AddressAutocompleteAction/useStoreAddress.ts
robinandeer: Hmm, maybe we can discuss the use of it at some point to come to a shared understanding of when it's useful etc.

To me it's simply a way to memoize the function definition so I don't think about it much anymore. It just feels safer to do it so that I don't accidentally depend on a non-memoized function in e.g. a `useEffect`-hook.

robinandeer:
Would need to redo this -- but will leave the PR open as a reference.

robinandeer:
This was a good attempt but we are gonna have to restart this one anyways

PR #534, opened 2021-06-29
Fix autocomplete content shift iOS

gustaveen:
## The issue
When clicking the `FakeInput` and triggering the `Modal`, the content shifts out of the viewport in iOS Safari.
This is due to the `FakeInput` getting focus when tapping it and the browser thinks the bottom of the page should be in view. We're unable to fool Safari by setting the focus manually to the input in the `Modal`.

https://user-images.githubusercontent.com/6661511/123826409-07dbba00-d900-11eb-8a5c-937a6298e59e.MP4

## Possible solution (Updated)
Make the `FakeInput` a regular span instead to remove all events attached to the input that causes iOS Safari to focus the viewport to the lower bottom of the screen. Wrap it in a button, because that's it kind of what it is anyway, a toggle for the Modal. The drawback is that it requires the user to tap the input in the `Modal`, but that's pretty much what Apple wants you to do 🙃

https://user-images.githubusercontent.com/6661511/124129562-2f589100-da7e-11eb-86c3-03bed5caf3d1.mov

### DIFF
--- a/src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
+++ b/src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
@@ -61,14 +61,13 @@ const Button = styled.button`
   }
 `
 
-const FakeInput = styled.span<{ fakePlaceholder: boolean }>`
+const FakeInput = styled.span<{ isMuted: boolean }>`
   display: block;
   max-width: 100%;
   padding: 0 16px;
   font-size: 20px;
   line-height: 1.5;
-  color: ${(props) =>
-    props.fakePlaceholder ? colorsV3.gray500 : colorsV3.gray900};
+  color: ${(props) => (props.isMuted ? colorsV3.gray500 : colorsV3.gray900)};
   text-overflow: ellipsis;
   white-space: nowrap;
   overflow: hidden;
@@ -316,7 +315,7 @@ export const AddressAutocompleteAction: React.FC<AddressAutocompleteActionProps>
           {props.tooltip ? <Tooltip tooltip={props.tooltip} /> : null}
           <Button onClick={handleButtonClick}>
             <FakeInput
-              fakePlaceholder={confirmedAddressLine === '' || searchTerm === ''}
+              isMuted={confirmedAddressLine === undefined && searchTerm === ''}
             >
               {fakeInputText}
             </FakeInput>
--- a/src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
+++ b/src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
@@ -59,20 +59,27 @@ const Button = styled.button`
   @media (min-width: 600px) {
     padding-bottom: 24px;
   }
+
+  & > input[disabled]::placeholder {
+    -webkit-text-fill-color: ${colorsV3.gray500};
+  }
 `
 
 const FakeInput = styled(Input)`
+  max-width: 100%;
   margin-left: 0;
   margin-right: 0;
   padding: 0 16px;
   line-height: 1;
   font-size: 48px;
   color: ${colorsV3.gray900};
-
-  max-width: 100%;
   text-overflow: ellipsis;
   white-space: nowrap;
   overflow: hidden;
+  &:disabled {
+    -webkit-text-fill-color: ${colorsV3.gray900};
+    opacity: 1;
+  }
 
   @media (min-width: 600px) {
     margin-top: 20px;
--- a/src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
+++ b/src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
@@ -74,10 +74,6 @@ const FakeInput = styled.span<{ fakePlaceholder: boolean }>`
   overflow: hidden;
   cursor: text;
 
-  ::placeholder {
-    -webkit-text-fill-color: ${colorsV3.gray500};
-  }
-
   @media (min-width: 600px) {
     padding: 0 32px;
     font-size: 48px;
--- a/src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
+++ b/src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
@@ -55,13 +55,10 @@ const Button = styled.button`
   outline: none;
   background: none;
   appearance: none;
+  cursor: text;
 
   @media (min-width: 600px) {
     padding-bottom: 24px;
-  }
-
-  & > input[disabled]::placeholder {
-    -webkit-text-fill-color: ${colorsV3.gray500};
   }
 `
 
@@ -76,9 +73,15 @@ const FakeInput = styled(Input)`
   text-overflow: ellipsis;
   white-space: nowrap;
   overflow: hidden;
+  cursor: text;
+
   &:disabled {
     -webkit-text-fill-color: ${colorsV3.gray900};
     opacity: 1;
+
+    ::placeholder {
+      -webkit-text-fill-color: ${colorsV3.gray500};
+    }
   }
 
   @media (min-width: 600px) {
--- a/src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
+++ b/src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
@@ -41,7 +41,6 @@ const Container = styled.div`
   flex-direction: column;
   align-items: center;
   justify-content: flex-start;
-  width: 100%;
 `
 
 const Card = styled(BaseCard)`
@@ -50,7 +49,7 @@ const Card = styled(BaseCard)`
 
 const Button = styled.button`
   width: 100%;
-  padding: 0 0 16px 0;
+  padding: 16px 0;
   border: none;
   outline: none;
   background: none;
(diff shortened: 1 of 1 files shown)


src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -40,15 +40,27 @@ const Container = styled.div`
...
-  padding-bottom: 24px;
+  @media (min-width: 600px) {
+    padding-bottom: 24px;
+  }
 
-  @media (max-width: 600px) {
-    padding-bottom: 16px;
+  & > input[disabled]::placeholder {
+    -webkit-text-fill-color: ${colorsV3.gray500};
   }
```
robinandeer, line 64: Can't this be applied to the "FakeInput" directly?

robinandeer:
Should we still try to mimic an input and add:

```
cursor: copy;
```

...to the button and the disabled input?

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
gustaveen: Yes! I was trying that unsuccessfully 😄 From what I know it should be like this:
```
&:disabled::placeholder {
```
But that didn't apply, am I writing it the wrong way? 🤔

gustaveen:
> Should we still try to mimic an input and add:
> 
> ```
> cursor: copy;
> ```
> 
> ...to the button and the disabled input?

Ah, what cursor does that render? A copy arrow? Maybe setting it to a text cursor would work as well?

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -40,15 +40,27 @@ const Container = styled.div`
...
   width: 100%;
+`
+
+const Button = styled.button`
+  width: 100%;
+  padding: 0 0 16px 0;
+  border: none;
+  outline: none;
+  background: none;
+  appearance: none;
```
styrelius, line 56: Maybe add `text` cursor to the `Button` and `FakeInput` to keep the feeling of it being an input? 🤔 
```suggestion
  appearance: none;
  cursor: text;
```

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -58,6 +70,12 @@ const FakeInput = styled(Input)`
   padding: 0 16px;
   line-height: 1;
   font-size: 48px;
+  color: ${colorsV3.gray900};
```
styrelius, line 73:
```suggestion
  color: ${colorsV3.gray900};
  cursor: text;
```

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -58,6 +70,12 @@ const FakeInput = styled(Input)`
...
+
+  &:disabled {
+    -webkit-text-fill-color: ${colorsV3.gray900};
+    opacity: 1;
```
styrelius, line 77: This seems to fix the disabled placeholder color?
```suggestion
    opacity: 1;
    ::placeholder {
      -webkit-text-fill-color: ${colorsV3.gray500};
    }
```

styrelius (commented):
This seems like a reasonable solution yes 👍

styrelius:
> Ah, what cursor does that render? A copy arrow? Maybe setting it to a text cursor would work as well?

I added suggestions for this before seeing that you've discussed it, but yes I think the `text` cursor would be nice 😀 Now it just renders the default cursor

robinandeer:
@gustaveen yes I meant to write "text" as the cursor type of course! And like @styrelius has updated I think we need it both on the button and input to make it seamless 👌

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
robinandeer: That looks correct to me... anyways, seems to work when nesting like this:

```javascript
const FakeInput = styled(Input)`
  &:disabled {
    -webkit-text-fill-color: ${colorsV3.gray900};
    opacity: 1;

    ::placeholder {
      -webkit-text-fill-color: ${colorsV3.gray500};
    }
  }
`
```

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
gustaveen: It's so weird, I tried that, but I can see that it updates the placeholder either. Does it work for you?

gustaveen: This is the downside of using a disabled input, since we need to set the opacity and add -webkit-text-fill-color to style the placeholder and input text in safari. According to mozilla it's not optimal https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-fill-color

I think I can live with that but it might be better to use a standard div instead, WDYT?

robinandeer: Yeah, most likely it's a better option long term. I thought it would be the easiest way to inherit all the styles instead of redefining them but it might turn out not to be the case!

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
gustaveen: Haha yup now it worked somehow, thank you! 🙌 Might be the port forwarding to the phone didn't update it properly before 🤪

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
styrelius: I'm fine either way, that 👆 works in Safari as well (both desktop and mobile) with Storybook at least

styrelius (dismissed):
Great job with this @gustaveen ! 🥳 🌤️

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -298,19 +314,16 @@ export const AddressAutocompleteAction: React.FC<AddressAutocompleteActionProps>
...
-            size={
-              (confirmedAddressLine || searchTerm || props.placeholder).length
-            }
-          />
-          {confirmedPostalLine ? (
-            <PostalAddress>{confirmedPostalLine}</PostalAddress>
-          ) : null}
+          <Button onClick={handleButtonClick}>
+            <FakeInput
+              fakePlaceholder={confirmedAddressLine === '' || searchTerm === ''}
```
robinandeer, line 319: `confirmedAddressLine` will never be an empty string... Shouldn't it be like:

```suggestion
              fakePlaceholder={confirmedAddressLine === undefined && searchTerm === ''}
```

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -282,6 +297,7 @@ export const AddressAutocompleteAction: React.FC<AddressAutocompleteActionProps>
     ? formatAddressLines(confirmedAddress)
     : []
 
+  const fakeInputText = confirmedAddressLine || searchTerm || props.placeholder
```
robinandeer, line 300: Perhaps this can be moved inline to the JSX since it's only used once?

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -45,33 +45,44 @@ const Container = styled.div`
...
 `
 
-const FakeInput = styled(Input)`
-  margin-left: 0;
-  margin-right: 0;
-  padding: 0 16px;
-  line-height: 1;
-  font-size: 48px;
-
+const FakeInput = styled.span<{ fakePlaceholder: boolean }>`
```
robinandeer, line 64: Another way to name the prop would be "muted". But either way works, it's only an internal API to the component!

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -45,33 +45,44 @@ const Container = styled.div`
...
+  outline: none;
+  background: none;
+  appearance: none;
+  cursor: text;
 
-  @media (max-width: 600px) {
-    padding-bottom: 16px;
+  @media (min-width: 600px) {
+    padding: 24px 0;
   }
```
robinandeer, line 61: I like this better. I used "max-width" to align with the rest of the code base. But perhaps it's time we move towards a more standard approach to responsive design!

robinandeer (commented):
Looks good! Just have a look at the comment about how to determine "fakePlaceholder" prop!

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
styrelius: Oh, missed this but I think since it's a boolean it would be nice to follow the convention to prefix whatever word you want with is or has (or can or should in special cases but you know what I mean)

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
gustaveen: Ah yes, good catch, thank you! 🙏

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
gustaveen: Hehe yes, slowly transforming it one media query at the time! 😄

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
gustaveen: Oh, that's a nice name! 👌

robinandeer (dismissed):
Great work! 🧹

src/Components/Actions/AddressAutocompleteAction/AddressAutocompleteAction.tsx
```diff
@@ -61,14 +61,13 @@ const Button = styled.button`
...
 `
 
-const FakeInput = styled.span<{ fakePlaceholder: boolean }>`
+const FakeInput = styled.span<{ muted: boolean }>`
```
styrelius, line 64: So maybe `isMuted` then? 😀

gustaveen: Sure! 👍

yalc.lock
```diff
@@ -0,0 +1,4 @@
+{
+  "version": "v1",
+  "packages": {}
+}
```
robinandeer, line 4: @gustaveen what is this file? 😅

gustaveen: oops, a bit to reckless with `git add .` 😂

robinandeer (approved):
Seems to work great!

PR #535, opened 2021-07-01
AddressAutocomplete cancel button cursor

styrelius:
### What?

Minor changes to the dismiss button in `AddressAutocomplete.tsx`:
- Name changed from `ModalHeaderButton` to `DismissButton` (since that's the only thing it's used for)
- 'pointer' cursor
- A slightly thinner outline when the button is focused, and some padding :)

**Focused cancel button:**
image

### DIFF
--- a/src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
+++ b/src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
@@ -43,11 +43,13 @@ const ModalHeaderLabel = styled.label`
 
 const ModalHeaderButton = styled.button`
   border: 0;
-  background: transparent;
+  background: ${colorsV3.gray300};
+  border-radius: 4px;
   appearance: none;
   outline: 0;
   position: relative;
   z-index: 2;
+  cursor: pointer;
 
   color: ${colorsV3.black};
   font-family: ${fonts.FAVORIT};
@@ -55,7 +57,7 @@ const ModalHeaderButton = styled.button`
 
   &:focus {
     border-radius: 4px;
-    box-shadow: 0 0 0 3px ${colorsV3.purple300};
+    box-shadow: 0 0 0 2px ${colorsV3.purple300};
   }
 `
 
--- a/src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
+++ b/src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
@@ -41,7 +41,7 @@ const ModalHeaderLabel = styled.label`
   font-weight: bold;
 `
 
-const ModalHeaderButton = styled.button`
+const DismissButton = styled.button`
   border: 0;
   background: ${colorsV3.gray300};
   border-radius: 4px;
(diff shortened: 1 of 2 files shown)


src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
```diff
@@ -41,21 +41,24 @@ const ModalHeaderLabel = styled.label`
...
   position: relative;
   z-index: 2;
+  cursor: pointer;
 
   color: ${colorsV3.black};
   font-family: ${fonts.FAVORIT};
   font-size: 16px;
 
   &:focus {
     border-radius: 4px;
```
robinandeer, line 60: Already defined as the base styles.

src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
```diff
@@ -41,21 +41,24 @@ const ModalHeaderLabel = styled.label`
...
+const DismissButton = styled.button`
   border: 0;
   background: transparent;
+  padding: 4px 2px;
+  border-radius: 4px;
   appearance: none;
   outline: 0;
   position: relative;
   z-index: 2;
+  cursor: pointer;
```
robinandeer, line 53: I never really got to the bottom of this but from W3 guidelines:

> The 'cursor' property
> pointer
>   The cursor is a pointer that indicates a link.

https://www.w3.org/TR/CSS21/ui.html#propdef-cursor

robinandeer (commented):
Nice improvements - just have a look at the cursor-comment. Perhaps we should discuss how it should work across the projects :-)

src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
styrelius: Wow, yes, I'm tired 😅

src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
styrelius: Oh, really sorry - I thought you just missed this and reasoned it was fastest just fixing it myself. But that is super weird IMO, I had no idea this was the case. Still, isn't the de facto convention still to use the `pointer` cursor when you want to indicate that something is clickable? To me I'm not sure it matters what the W3 guidelines are in this case, since the convention feels so established? Not using the pointer for a button really looks like a mistake to me and it can't just be me? Or?

robinandeer: Yeah, you are right about that. Let's keep it "clickable" :-)
The usage has probably changed since the original draft of that document.

robinandeer (approved):
Nice changes!

src/Components/Actions/AddressAutocompleteAction/AddressAutocomplete.tsx
gustaveen: Ah interesting, haven't reflected on that really. Am so used to buttons having `pointer` as well. So I'm good with that 😄

gustaveen (approved):
Nice touch up! 👯

PR #545, opened 2021-07-08
Update card link color

gustaveen:
## Why
Contrast was quite poor so switched to a darker swatch and increased the font size since it's very tiny

#### Before
Screenshot 2021-07-08 at 16 29 43

#### After
Screenshot 2021-07-08 at 16 29 53

### DIFF
--- a/src/Components/Actions/SelectAction/SelectAction.tsx
+++ b/src/Components/Actions/SelectAction/SelectAction.tsx
@@ -6,7 +6,7 @@ import { callApi } from '../../API'
 import { ApiContext } from '../../API/ApiContext'
 import { Loading } from '../../API/Loading'
 
-export type SelectActionProps = {
+type SelectActionProps = {
   passageName: string
   action: any
   changePassage: (name: string) => void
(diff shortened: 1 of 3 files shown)


robinandeer (dismissed):

src/Components/Actions/SelectAction/SelectAction.tsx
```diff
@@ -6,7 +6,7 @@ import { callApi } from '../../API'
 import { ApiContext } from '../../API/ApiContext'
 import { Loading } from '../../API/Loading'
 
-type SelectActionProps = {
+export type SelectActionProps = {
```
styrelius, line 9: Does this need to be exported? Doesn't seem to be used anywhere 🤔

src/Components/Actions/SelectAction/SelectOption.stories.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { Story, Meta } from '@storybook/react/types-6-0'
```
robinandeer, line 2: That's an interesting import path 😅

robinandeer (approved):
Nice update!

PR #549, opened 2021-07-09
Fix 'confirmSuggestion' function in AddressAutocompleteAction.tsx

robinandeer:
Update function to confirm a selected address suggestion.
Make sure to always send the property ID instead of the access ID for houses.

...also adds a missing semicolon in emotion component.

## Why

The address for a house has two separate IDs - access and property. We always want to send the property ID to the API. There's no obvious way to tell them apart. Therefore we accidentally sent the access ID.

After this update: to verify a selected house address we:
1. make a second call to the API
2. compare that we only receive a single suggestion for that street address
3. assuming that this suggestion includes the property ID we assign that as the "confirmed address"
4. when user clicks "continue" we submit it to the store


gustaveen (approved):
🙌

PR #550, opened 2021-07-13
Update example command to push tags

robinandeer:
Found a way to push commit and tag in one command!

Also - this is not recommended since you can end up pushing unrelated tags even if you didn't intend to:

```
git push --tags
```

Reference: https://stackoverflow.com/a/26438076


gustaveen (approved):
🙌

PR #561, opened 2021-07-27
APP-606: Use passage-level API call for `houseInformation` instead of `action`-level

hedvigoscar:
I think this might just work well enough 🤔


hedvigoscar:
APP-606

hedvigoscar:
Obsoleted by https://github.com/HedvigInsurance/android/pull/648

PR #564, opened 2021-08-06
Angel app metadata for DK

axelbacklund:
Needed for the app so we can switch to native Embark for DK.


robinandeer (approved):
We don't parse any of this data (yet?) on the web so looks good to me!

PR #565, opened 2021-08-06
Add possibility to check null/undefined values

robinandeer:
## What?

- Explicitly parse null as a value in expressions.
- Add tests for parsing null in expressions.
- Add tests for comparing equals/not equals against null values.
- Use triple equals when checking expression types. <= this is just clean-up

## Why?

I've noticed a need to check against the existence of values in the Embark store when using redirects.

The specific example is for handling underwriting guideline hits when creating quotes. The only way for Embark to know is whether or not it receives some GraphQL results. There's no specific errors coming from Giraffe.

## Demo

```
<Redirect to="[[errorPath]]" when="someErrorKey != null">
</Redirect>
```


src/Parsing/parseStoryData.test.ts
```diff
@@ -0,0 +1,24 @@
+import { parseExpression } from './parseStoryData'
+
+test('parse equality expression with string value', () => {
+  const expression = parseExpression('someKey == hello world')
+  expect(expression?.type).toBe('EQUALS')
+  // @ts-ignore
+  expect(expression?.key).toBe('someKey')
+  // @ts-ignore
```
robinandeer, line 8: Based on how the `Expression` type is defined, there's no way to narrow down the type to show it should have a `value` field....

src/Parsing/parseStoryData.ts
```diff
@@ -555,7 +555,7 @@ interface BinaryExpression {
   __typename: string
   type: string
   key: string
-  value: string
+  value: string | null
```
robinandeer, line 558: Had to add this -- but we seem to through away this type information because it's redefined again in https://github.com/HedvigInsurance/embark/blob/master/src/Utils/ExpressionsUtil/expression.ts#L14 😛

hedvigoscar (approved):
Looks good to me! I thought we had this already, to be honest 🤔

robinandeer:
Thanks @hedvigoscar - seems like the comparison value is never evaluated to anything but a string. We are doing the "loose" comparison with `==` instead of `===` but I checked and it doesn't make a difference.

hedvigoscar:
Interesting. Sounds like that might be based on per-client implementation, we should check our implementations on iOS and android as well.

robinandeer:
Yes indeed, it's easy to forget all clients. Been wondering if there's a way for Embark to handle more of that logic on the server 🤔

sampettersson (approved):
Makes sense, iOS already did this 🤸

PR #566, opened 2021-08-06
Update to Web Onboarding DK - Contents With Autocomplete => handle UW hit error

robinandeer:
## What?

Check `__typename` value after calling `createQuote` to determine whether we got an underwriting guidelines hit.

Redirect the user to the appropriate passage.

## Why?

Now the user just get's a white screen if there's a UW hit.

References: [GRW-268]

[GRW-268]: https://hedvig.atlassian.net/browse/GRW-268


hedvigoscar (dismissed):

styrelius (approved):
🥇

PR #569, opened 2021-08-10
Handle UW hit for house quotes in Web Onboarding - Swedish Needer

robinandeer:
Pick up UW hits when creating a house quote and redirect user to the proper error passage.

I renamed the passage `ApartmentUwlimitsHit` to `OfferUwlimitsHit` to reflect that it now handled UW hits for both house and apartment quotes.

The apartment flow also includes the `OfferSuccess` passage. However, there's no conflict adding the redirect here.


styrelius (commented):
👍 

Also, regarding the fact that the flow in English differs from the one in Swedish, I have no idea why that is. Sounds like a mistake to me 🤷‍♀️

styrelius (approved):

gustaveen:
🙌 
Hm that is weird, what is it that differ between them? @robinandeer

PR #579, opened 2021-09-02
Upgrade to Emotion 11

robinandeer:
## What

Upgrade to latest Storybook.
Update emotion babel plugin.
Import from @emotion/react instead of @emotion/core.
Add webpack workaround to make Storybook work with latest emotion.
Remove old emotion dependencies.

## Why

- If we want to use Emotion 11 we need to live with the workaround in `./storybook/main.js` until Storybook is updated.

- We need to update the same package in @HedvigInsurance/brand to match peer dependencies.


robinandeer:
I ran into a strange scenario where we have some unmet peer dependencies:

```
warning "@storybook/addon-essentials > @storybook/addon-measure@2.0.0" has unmet peer dependency "@storybook/components@^6.3.0".
warning "@storybook/addon-essentials > @storybook/addon-measure@2.0.0" has unmet peer dependency "@storybook/core-events@^6.3.0".
warning "@storybook/addon-essentials > @storybook/addon-measure@2.0.0" has unmet peer dependency "@storybook/theming@^6.3.0".
```

But when I install them I get this warning:

```
warning Pattern ["@storybook/components@^6.3.7"] is trying to unpack in the same destination "/Users/robinandeer/Library/Caches/Yarn/v6/npm-@storybook-components-6.3.7-42b1ca6d24e388e02eab82aa9ed3365db2266ecc-integrity/node_modules/@storybook/components" as pattern ["@storybook/components@6.3.7","@storybook/components@6.3.7","@storybook/components@6.3.7","@storybook/components@6.3.7","@storybook/components@6.3.7","@storybook/components@6.3.7","@storybook/components@^6.3.0","@storybook/components@6.3.7","@storybook/components@6.3.7"]. This could result in non-deterministic behavior, skipping.
```

Any idea what's up? Should we worry about it? I haven't explicitly added the peer dependencies.

gustaveen:
Nice! I'd like to use emotion 11 so I think it's a small trade off to use the workaround in the storybook setup. Hm, weird with the peer dependencies, sounds like it should have been solved from this conversation https://issuehunt.io/r/storybookjs/storybook/issues/15367 But can't see that it breaks anything though. 

We will need to update emotion in web-onboarding as well in order to bump version right?

package.json
```diff
@@ -90,10 +86,11 @@
     "@babel/plugin-proposal-optional-chaining": "^7.8.3",
     "@babel/preset-react": "^7.8.3",
     "@babel/preset-typescript": "^7.8.3",
-    "@storybook/addon-actions": "^6.1.11",
-    "@storybook/addon-essentials": "^6.1.11",
-    "@storybook/addon-links": "^6.1.11",
-    "@storybook/react": "^6.1.11",
+    "@emotion/babel-plugin": "^11.3.0",
+    "@storybook/addon-actions": "^6.3.7",
```
gustaveen, line 90: I think `addon-actions` can be removed since it's part of the essentials package now https://storybook.js.org/addons/tag/essentials/

robinandeer: Oh nice! Removed it :D

robinandeer:
Yeah we should update Emotion in web-onboarding as well as @hedviginsurance/brand I believe!

Hopefully the peer dependencies thing works itself out as we keep upgrading

gustaveen (approved):

gustaveen (approved):

gustaveen (approved):
Let's go! 🙌

package.json
```diff
@@ -90,10 +86,10 @@
...
     "@babel/preset-react": "^7.8.3",
     "@babel/preset-typescript": "^7.8.3",
-    "@storybook/addon-actions": "^6.1.11",
-    "@storybook/addon-essentials": "^6.1.11",
-    "@storybook/addon-links": "^6.1.11",
-    "@storybook/react": "^6.1.11",
+    "@emotion/babel-plugin": "^11.3.0",
+    "@storybook/addon-essentials": "^6.3.7",
+    "@storybook/addon-links": "^6.3.7",
+    "@storybook/react": "^6.3.7",
```
gustaveen, line 92: As always with package upgrade PRs, there's a new version before you have time to merge 😂 Look like `storybook` is at `6.3.8` now @robinandeer

robinandeer: haha true, but we only need to run "yarn upgrade" I guess since it will still resolve to the latest version :)

PR #580, opened 2021-09-02
Upgrade to latest TypeScript

robinandeer:
## What

Update to official rollup typescript plugin.

## Notes

The update requires a fix in Storybook so that's why this branch is on top of #579


src/Components/templateExpression.ts
```diff
@@ -132,10 +132,9 @@ const parseAbstractExpression = (
         }
         if (abstractExpression) {
           if (abstractExpression.type === 'binaryOperator') {
-            abstractExpression = {
-              ...abstractExpression!,
+            abstractExpression = Object.assign(abstractExpression, {
```
styrelius, line 135: Why is this changed?

styrelius (commented):
Nice! 🏗️ Just a question

src/Components/templateExpression.ts
robinandeer: Oh yeah, I only added the info in the commit message:

Use Object.assign instead of spread operator because of this issue: microsoft/TypeScript#10727.

robinandeer:
Oh dammit, now I duplicated some commits... but the changes should still only be about TS

PR #583, opened 2021-09-02
Feature/move embark schema to giraffe

cpiehl1:
Adding endpoints for resolving the queries from giraffe instead, to avoid schema stitching.


styrelius (commented):
I think this probably makes a lot of sense (at least the avoiding schema stitching part) but I'm confused about that we also have the `/angel-data` route (mostly because I don't know how it works today). Are both needed?

vonElfvin (approved):
Very neat and clean!

cpiehl1:
> I think this probably makes a lot of sense (at least the avoiding schema stitching part) but I'm confused about that we also have the `/angel-data` route (mostly because I don't know how it works today). Are both needed?

I added both because they are present in the schema and and have quite different resolvers, one returns `embarkStory` and the other `[embarkStoryMetadata]`. I am not sure if both are actually used but I wanted to keep the API identical in giraffe.

PR #584, opened 2021-09-02
Upgrade Node.js to latest LTS

robinandeer:
LTS = long term support.
Upgrade base Docker image.
Upgrade Node types.


robinandeer:
Hmm it fails because the pipeline uses the old Node.js version. How do we handle such a case? The pipeline is defined inline so it will affect all PRs 🤔 

https://g.codefresh.io/pipelines/edit/workflow?id=5faa688b09f3f88dcd57303c&pipeline=embark&projects=Web&projectId=5fa3c2982d5cdeef9964a26a&rightbar=steps

gustaveen (approved):
🙌

PR #585, opened 2021-09-03
Use swedishHouse and swedishApartment quote types

tobiasbexelius:
Their deprecated counterparts 'house' and 'apartment' will be removed shortly


sampettersson (approved):
We would also need to check if there is any `GraphQLAPI` usages in the angel flows that uses `apartment/house` this implementation is the old API implementation which is used in Swedish flows.

PR #590, opened 2021-09-07
Add new VariantedOfferRedirect

sampettersson:
> This tells the consumer the ids of all quotes + the ids of the quotes that should actually be `selected` when they land the offering screen, this is to support the toggling functionality on the offer page.

Not super sure if we should add a completely new `OfferRedirect` or just smash this into the existing one 🤔


gustaveen:
Cool! I think it makes sense to create a new one for starters 👍

gustaveen:
This looks good @sampettersson and makes sense to pass selected keys 🙌 What I wonder is how we should choose what id's to pass in? We basically have the first passage with two options to work with :) 
Screenshot 2021-09-17 at 11 42 00

Should we in that step create the ids (with a GeneratedVariable) and store them? And create property for selected keys as well? If you understand what I mean :)

gustaveen (approved):
🙌

PR #602, opened 2021-09-20
Update to Web Onboarding SE - Needer

gustaveen:
## What
Fix `MultiactionVariable` for `extraBuildings`. Nested Variables was missing the `from` prop and was therefore not collected from the store.

```
<MultiactionVariable key="extraBuildings">
    <Variable key="type"
              from="type"
              as="string"
    >
    </Variable>
    <Variable key="area"
              from="area"
              as="int"
    >
    </Variable>
    <Variable key="hasWaterConnected"
              from="hasWaterConnected"
              as="boolean"
    >
    </Variable>
  </MultiactionVariable>
```


mehdielket:
Could you please clarify what the problem was (from a user perspective) ?

sampettersson (approved):

gustaveen:
> Could you please clarify what the problem was (from a user perspective) ?

The extra building values was not collected from the store and was set to null, making the mutation fail and the user getting stuck on the error page :) @mehdielket

PR #614, opened 2021-09-28
Add expression support to VariantedOfferRedirect

gustaveen:
## What 
Add support for `when` prop to `VariantedOfferRedirect`. This follows the pattern we have for a regular `Redirect`.

## Why
We need to conditionally pass `selectedIds` to know what bundle the user have picked. In the first solution we used a passage with a regular `<Redirect>` followed by a passage with a `<VariantedOfferRedirect>`. This wouldn't trigger the last redirect automatically plus it feels a bit repetitive to have two redirects in a row.  

Things to consider: 
`VariantedOfferRedirect` needs to have an expression (a `when` prop) in order to be executed. The thinking is that you would just use `OfferRedirect` if you just want to pass a single bundle.

**Implementation**
```
<VariantedOfferRedirect when="insuranceType == 'homeAccident'"
                        keys="swedishHomeQuoteId,swedishAccidentQuoteId"
                        selectedkeys="swedishHomeQuoteId,swedishAccidentQuoteId"
>
</VariantedOfferRedirect>

<VariantedOfferRedirect when="true"
                        keys="swedishHomeQuoteId,swedishAccidentQuoteId"
                        selectedkeys="swedishHomeQuoteId"
>
</VariantedOfferRedirect>
```

### Before
Screenshot 2021-09-28 at 10 47 08

### After
Screenshot 2021-09-28 at 10 37 02


sampettersson (approved):
Reasonable 😄

PR #619, opened 2021-09-29
Update to accident-cross-sell-SE

zakkjellin:
Removed a space and moved a message to another passage


mehdielket:
I assume this is not relevant anymore? Shall we close it?

PR #632, opened 2021-10-11
Add audio recorder action

sampettersson:


schema.ts
```diff
@@ -381,16 +381,26 @@ const typeDefs = /* GraphQL */ `
         data: EmbarkSwitchActionData!
     }
 
+    type EmbarkAudioRecorderActionData {
+        key: String!
```
hedvigoscar, line 385: file upload URL?

hedvigoscar (approved):
lgtM

schema.ts
sampettersson: yeah, i was thinking the frontend component would upload it and store the url :)

hedvigoscar (approved):

PR #669, opened 2021-11-01
GRW-482 / Fix / Missing error messages in Web Onboarding SE - Needer

robinandeer:
## What

Fix spelling of "Messeage" => "Message" in a few elements.

## Why

Error messages are not displaying for error passages in SE flow.

## Demo

It's working now:

Screenshot 2021-11-01 at 13 53 50

## For future reference

It's possible to find out that you've correctly spelled the element name in the preview. But it should be more obvious.

<table>
<tr>
	<td>Incorrect
	<td>Correct
<tr>
	<td>Screenshot 2021-11-01 at 13 51 41
	<td>Screenshot 2021-11-01 at 13 51 50
</table>


gustaveen (approved):
Thank you for finding that! :D 
giphy (5)

PR #672, opened 2021-11-02
Fix schema for VariantedOfferRedirect

sampettersson:
GraphQL Schema wasn't up to date with the expression functionality :)


robinandeer (approved):
Haven't tried locally but code checks out to me 😄

PR #674, opened 2021-11-03
GRW-503 / Pass currentInsurer + dataCollectionId (Web Onboarding SE - Switcher Without Accident)

robinandeer:
Horrible diff to review but I basically just added the above parameters to the quote mutation.


gustaveen (approved):
🙌

PR #679, opened 2021-11-03
Use `HouseInformation` query in `manualAddresVilla` passage (Web Onboarding SE - Needer)

robinandeer:
We used a `PersonalInformation` query here but that seems incorrect based on the previous setup.

This logic looks correct now, right?

Screenshot 2021-11-03 at 13 36 01

Both "manualAddressVilla" and "newAddressVilla" should have the same next steps, right?


sampettersson (approved):

gustaveen (approved):
🙌

PR #680, opened 2021-11-03
Remove references to deprecated components

robinandeer:
Remove HouseInformationAPI + PersonalInformationAPI components and related code.

We should use `GraphQLAPI` component for these kinds of things moving forward.
@sampettersson has already confirmed that Apps are not using these.

See #667.

> FYI 1: Not quite ready to be merged - we need to switch to the new embark flows in SE first :)

> FYI 2: Is it me or why does prettier want to re-format a bunch of stuff?


schema.ts
```diff
@@ -108,28 +108,6 @@ const typeDefs = /* GraphQL */ `
         data: EmbarkApiGraphQLMutationData!
     }
 
-    type EmbarkApiPersonalInformationData {
-        match: EmbarkLink!
-        noMatch: EmbarkLink!
-        error: EmbarkLink!
-    }
-
-    type EmbarkApiPersonalInformation implements EmbarkApiCore {
```
sampettersson, line 117: Lets see if this works or breaks apps, I don't think we fetch it explicitly but it might break, I don't think so, in that case we need to keep them and mark them deprecated. Lets confirm after this is merged 😄

robinandeer: Good point. Can that be tested before staging?

Is there "review apps" for Giraffe? 😅

sampettersson: Hmm, no 😓

robinandeer: Oki, well we can always do some testing directly after merging and then re-introduce the changes if they cause issues

sampettersson: Oui

robinandeer:
This one is out of date -- will reopen with the same changes at a later occasion.

PR #681, opened 2021-11-03
GRW-533 / update animations

dendoesit:
What?

Updated animation speed.

Why? 

They were too slow.

Now

https://user-images.githubusercontent.com/63097139/140067871-054fa9f1-49e9-4ec8-9553-0c953e8e145e.mp4


robinandeer (dismissed):
The video demo looks very distorted on my computer but from what I can tell you've fixed the animation!

gustaveen:
Thanks for taking another stab at this Dan 🙌 
I played around a bit locally but I can't seem to change the transition, tried to make it even quicker. I would assume that setting the duration to 0.1 seconds would make it a lot faster but the timing seems to be exactly the same? 🤔
Do you know why? Seems like it doesn't change at all for me @danpopoutanu @robinandeer 

https://user-images.githubusercontent.com/6661511/140124222-380f6bea-8abd-4a06-b192-312a7e43bc8b.mov

dendoesit:
@gustaveen you can just change the stiffness of the spring animation. 
Higher values will create more sudden movement. Set to 100 by default.
You can also add velocity.

gustaveen:
> @gustaveen you can just change the stiffness of the spring animation. Higher values will create more sudden movement. Set to 100 by default. You can also add velocity.
> 
> Bumped it to 500 and removed duration so it doesn't lag behind. https://user-images.githubusercontent.com/63097139/140173157-a431b490-0a6e-4f94-9fed-d46bcff87687.mp4

Ah, thank you for that clarification! Saw this in the docs and it figures: "Note: duration and bounce will be overridden if stiffness, damping or mass are set"

This looks snappy and nice 🙌 I think we can even bring it down a tiny bit to 450 in stiffness :)

dendoesit:
Updated to 450. 

https://user-images.githubusercontent.com/63097139/140278501-254d4110-a0f9-4e41-964c-cf06590c6b2a.mp4

robinandeer (approved):
Cool stuff!

gustaveen (approved):
Awesome Dan! :raised_hands:

PR #682, opened 2021-11-03
Handle number being passed to `NumberEditCard`

robinandeer:
## What

Handle number being passed to the value prop in `NumberEditCard`.

## Why

It failed since `.trim()` doesn't exist on number in JS.
Number was being passed from Embark store.


styrelius (approved):
🕵️‍♂️

src/Components/Actions/NumberActionSet/NumberEditCard.tsx
```diff
@@ -108,7 +108,7 @@ export const NumberEditCard: React.FC<NumberEditCardProps> = (props) => {
     }
   })
 
-  const shouldPushUp = !isSm || (props.value || '').trim().length > 0
+  const shouldPushUp = !isSm || String(props.value || '').trim().length > 0
```
guilhermespopolin, line 111: 🤔 So if `0` is passed as `prop.value` we should consider as nothing was passed?

guilhermespopolin (approved):
Is this a good opportunity to add a story for this component?

src/Components/Actions/NumberActionSet/NumberEditCard.tsx
robinandeer: Hmm that is a good point. Should probably be "??", right?

guilhermespopolin: If I understood correctly, `shouldPushUp` should be `true` if the small breakpoint was not match and we get *something* as a value (`props.value`). If that's correct, any `string` our `number` values are valid. Since we're expecting a `string` | `number` based on type definition and therefore nullish values are not expected (`null` and `undefined`), I think we should keep it as:

```suggestion
  const shouldPushUp = !isSm || String(props.value).trim().length > 0
```

Does that make sense? 🤔

robinandeer: The thing is that we pass an "any" value from the Embark store so we lose the type safety unfortunately. That's why I don't want to make to make too many assumptions here.

PR #683, opened 2021-11-03
 updated jsdom + pretty-quick

dendoesit:
What ? 

Updated jsdom + pretty-quick to latest versions

Why ?

Monthly updates


guilhermespopolin (approved):

robinandeer (approved):
Nice update! No breaking changes to care about? I think it's great if we gather any relevant links to release notes in these PR we find doing the research :)

dendoesit:
@robinandeer I think that is a good idea. 

JSDOM:
Requires node v12
Fixed SSL certificate checking for WebSocket connections. Previously, invalid SSL certificates were always accepted; now, they properly respect the ResourceLoader's strictSSL option (which defaults to true).
Changed the global in which almost all Promise and TypeError instances are created to be the jsdom global, not the Node.js global. This could affect any code that uses instanceof.

Pretty-quick:
BREAKING CHANGE: require node 10 and prettier 2

PR #687, opened 2021-11-04
GRW-540 / Fix button color in iOS 15 (previous insurance provider)

robinandeer:
...plus finally got around to fixing the design bug in "OtherButton" :relieved:

Screenshot 2021-11-04 at 15 08 31


gustaveen (approved):
YES!

PR #689, opened 2021-11-05
Fix extra building water connected translation

gustaveen:
## What

- "Indraget vatten" was a hard coded string. Translate label instead.
- Fix som React key warnings.

Also, husky run prettier on `parseStoryData.ts` so kind of hard to see the small diff. I would like to get rid of that hook, we should just run format on save anyways. 

Screenshot 2021-11-05 at 15 13 36

https://hedvig.atlassian.net/browse/GRW-556?atlOrigin=eyJpIjoiODBlMzNkMGJkM2IzNDhhYmE1NmM1NTU3NDZiNTc1NmEiLCJwIjoiaiJ9


src/Parsing/parseStoryData.ts
```diff
@@ -226,9 +225,9 @@ const getDropdownAction = (
   }
 }
 
-const getSwitchAction = (switchActionNode: Element) => {
+const getSwitchAction = (switchActionNode: Element, translate: Translator) => {
   const key = switchActionNode.getAttribute('key')
-  const label = switchActionNode.getAttribute('label') || ''
+  const label = translate(switchActionNode.getAttribute('label') || '')
```
gustaveen, line 230: This is the diff for getSwitchAction

src/Components/Actions/InlineActions/InlineDropdownAction.tsx
```diff
@@ -62,8 +62,10 @@ export const InlineDropdownAction = (props: DropdownActionProps) => (
       <option value="" disabled selected>
         {props.label}
       </option>
-      {props.options.map((option) => (
-        <option value={option.value}>{option.text}</option>
+      {props.options.map((option, index) => (
+        <option key={index} value={option.value}>
```
robinandeer, line 66: nit: Isn't it better to use `option.value` as the key? Should be unique.

robinandeer (approved):
Looks good :)

src/Components/Actions/InlineActions/InlineDropdownAction.tsx
gustaveen: True 👍 I think it's fine though to use index as long as it's a static list and not being reordered :)

PR #690, opened 2021-11-05
GRW-537 / Fix safari select font issue

robinandeer:
Override styles only for safari browsers.

Not ideal since we can't use our custom font but I still think it's an improvement overall.

### Safari

Screenshot 2021-11-05 at 16 39 49

Screenshot 2021-11-05 at 16 39 55

### Chrome
Screenshot 2021-11-05 at 16 40 14

Screenshot 2021-11-05 at 16 40 20


styrelius (approved):
⛏

guilhermespopolin (approved):
image

PR #698, opened 2021-11-09
Update to Web Onboarding SE - Switcher - Update trackable properties

gustaveen:


gustaveen:
need more trackable properties

PR #703, opened 2021-11-09
Remove excess scrollbars

dendoesit:
What?

Hide scrollbars on desktop 

Why ?

We have double scrollbars because of overflow: scroll. 

Not sure why we needed overflow: scroll on the passage container as removing this makes the scrollbar show up when needed ( when the page is small and needs to be scrolled ).
Tested it on a couple of flows on different screen sizes. 

Before:

image

Now:

image
image


robinandeer:
Awesome that you are looking into this! Makes a big difference for users 😌

It's surprising that we can solve it without even "overscroll: auto" 😅 I guess you verified that the we don't scroll away the header, right?

guilhermespopolin (dismissed):

dendoesit:
@robinandeer  I guess it's mostly desktop users that are affected by this but still an improvement 🎉 

I'm not sure what you mean by scroll the header away because you could do that before.

robinandeer (changes requested):
@danpopoutanu what I mean is that the header should be fixed at the top of the page like this:

Screenshot 2021-11-10 at 17 48 34

When I try these changes on web-onboarding, I noticed that things simply get cut-off and I can't scroll to see the rest of the content:

Screenshot 2021-11-10 at 17 59 51

I think we have a problem that testing things in Angel doesn't seem to behave the same as on web-onboarding.

I think the solution we are after is `overflow-y: auto`. Can you verify that this works on windows/firefox?

src/Components/Passage.tsx
```diff
@@ -31,8 +31,8 @@ interface PassageProps {
 const ChatContainer = styled.div<{ hasHeader?: boolean }>`
   display: flex;
   height: ${({ hasHeader }) => (hasHeader ? 'calc(100% - 4.5rem)' : '100%')};
-  overflow: scroll;
   flex-direction: column;
+  overflow: auto;
```
robinandeer, line 35: Just because we don't intend to scroll in x-axis

```suggestion
  overflow-y: auto;
```

dendoesit: was just about to write.
 we have the extra-buildings screen where if you have more buildings on a small screen we need to scroll left or right.

robinandeer: Oh that's very annoying... should probably have another solution for that specifically but I will approve for now!

robinandeer (approved):
Nice change! Let's remember to fix the extra building setup so it can scroll horizontally independently.

src/Components/Passage.tsx
dendoesit: nevermind. overflow-y seems to work as well. the extra-building already has scroll implemented when there are multiple boxes. 

tested on several flows and seems to work.

robinandeer (approved):
If this works for extra buildings -- all good!

PR #704, opened 2021-11-12
Delete unused angel data

gustaveen:
## What 
Delete unused angel data for DK stories without address autocomplete and old stories for SE.

## Why
Not being used anymore. Please confirm though @sampettersson @tarikstafford @hedvigoscar @tarikstafford so I don't break shit for you :D


robinandeer (approved):
Awesome! Perhaps we should rename the "with autocomplete" stories also at some point

gustaveen:
> Awesome! Perhaps we should rename the "with autocomplete" stories also at some point

Indeed! @robinandeer I'm just wondering if that would be tricky to sync a release of that for the apps?

robinandeer:
Hehe yeah it's not such a big deal. We will also need new flows after the API migration 🙃

PR #706, opened 2021-11-16
Footer Text component

dendoesit:
What?

New Text component similar to message.
New field to be parsed.

Why ? 

We need it in the personummer screen.
Jira ticket : https://hedvig.atlassian.net/browse/GRW-407

- [x] Add markdown

image

with link 
image


src/Components/Passage.tsx
```diff
@@ -370,6 +372,9 @@ export const Passage = (props: PassageProps) => {
                 </HelpButton>
               </HelpButtonWrapper>
             )}
```
dendoesit, line 374: I added it in the Bottom Container for now. Not sure if there are any plans of reusing this.

robinandeer (commented):
I'm a little worried that using `Text` is a bit too generic for this component.

Since we are always rendering it at the bottom of the passage I feel like `FooterText` or similar would better communicate the purpose of it. Do you agree?

src/Components/Text/Text.tsx
```diff
@@ -0,0 +1,39 @@
...
+  margin-top: 10px;
+  max-width: 440px;
+  font-family: ${fonts.FAVORIT};
+
+  @media (min-width: 600px) {
+    margin-top: 24px;
+  }
+`
+
+export const TranslationNode: React.FC = ({ children }) => <>{children}</>
```
gustaveen, line 31: Where is this being used?

src/Components/Passage.tsx
gustaveen: That works good! I don't think there will be a any other case for this component, so I think we could call it Footer. It might be that we want to render more things in it than just text might be a better name or WDYT?

src/Components/Text/Text.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export type TextProps = {
+  content: Content
+}
+
+const TextContainer = styled.div`
+  text-align: center;
+  color: ${colorsV3.gray700};
+  font-size: 12px;
```
gustaveen, line 20: For now components we should aim to use rems for `font-size`, `margin`, `padding` etc :)

src/Components/Text/Text.stories.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import { Story, Meta } from '@storybook/react/types-6-0'
+import { Text, TextProps } from './Text'
+
+export default {
+  title: 'Components/Text',
+  component: Text,
+} as Meta
```
gustaveen, line 8: Might be nice to have it centered in the story :) 
```suggestion
export default {
  title: 'Components/Text',
  component: Text,
  parameters:{
    layout: 'centered'
  }
} as Meta
```

gustaveen (commented):
Nice one @dendoesit! The Markdown PR is merged so you could add that now 🙌

src/Components/Passage.tsx
dendoesit: I agree. The case where we might render more things inside of it makes a good argument for naming it Footer instead of FooterText

dendoesit:
Added markdown, fixed minor issues and updated component name. @gustaveen

src/Components/Footer/Footer.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const TextContainer = styled.div`
+  text-align: center;
+  color: ${colorsV3.gray700};
+  font-size: 0.75rem;
+  line-height: 1.33;
+  margin-top: 0.625rem;
+  max-width: 440px;
+  font-family: ${fonts.FAVORIT};
+
```
gustaveen, line 26: I think you can add the styling for links in here. If you preview it in Angel it will render as a a blue link :) 
Screenshot 2021-11-22 at 16 27 03

gustaveen: Let's check with Zak what purple it should have, it should be darker than pruple500, probably purple900
https://www.figma.com/file/xavk7WJVU1xqQKdJHqIlW6?node-id=440:5506#126827875

gustaveen (commented):
Great Dan! Just and update to the link styling and color

src/Components/Footer/Footer.tsx
dendoesit: good spot. I thought that's the default color the link gets after it has been clicked. will confirm with Zac and change.

gustaveen (approved):

PR #707, opened 2021-11-18
Update to home-content-cross-sell-NO

hugokallstrom:
Duplicate of story "Web Onboarding - NO English Contents" - but used in the cross sell flow.


hedvigoscar (changes requested):
- is `isMover` and the `cost`-properties taken from the mutation result used for anything?
- The error-passage at the end is empty, should it have content?
- The flow is currently untranslated, should it be translated? 🙂

PR #709, opened 2021-11-18
Update to travel-cross-sell-NO

hugokallstrom:
Story used for cross selling norwegian travel insurance


hedvigoscar (commented):
This story re-uses the keys from the accident cross-sell story, is that appropriate or should we have separated keys?

hugokallstrom:
> This story re-uses the keys from the accident cross-sell story, is that appropriate or should we have separated keys?

Think it makes sense to use the same keys here

StylianosGakis (approved):

PR #710, opened 2021-11-22
Update to travel-cross-sell-NO-youth

hugokallstrom:


hedvigoscar (commented):
Is it appropriate to have an embark flow at all for this cross-sell? Wouldn't it be sort of a flash of the embark-screen, and then right away show the offer? 🤔

PR #711, opened 2021-11-23
Fix story setting homeAndAccident instead of homeAccident

sampettersson:
In the offer step homeAccident is expected and not homeAndAccident


robinandeer (approved):
Oh that's bad! Nice that you found it

I guess needer flow is okay?

sampettersson:
@robinandeer yep, needer flow was ok :)

robinandeer:
@gustaveen Did you see this? Perhaps also a reason why we are getting errors in production...

gustaveen:
Ah nice catch! 🙌 @sampettersson 
We don't use this in production @robinandeer :) We use `Web Onboarding SE - Switcher Without Accident` so this option isn't present.

PR #715, opened 2021-11-25
Add caching when loading stories from Embark

hedvigoscar:
My highly unscientific benchmarks observed speedups on the magnitude of 10-25x. (100ms-250ms => ~10ms) Nice!


hugokallstrom (approved):
LGTM!

PR #716, opened 2021-12-03
Grw 436/insurely

dendoesit:
##What ? 

Created the iframe as a separate component. Still has console logs and comments as the code will be revised when we push the final version of the iframe. 

This will eventually be put instead of ExternalInsuranceAPI.

As part of the POC there will be a separate flow created to test out the iframe flow. 

##Why ? 

POC for Insurely Iframe.

image


src/Components/Actions/InsurelyIframe/InsurelyIframe.tsx
```diff
@@ -0,0 +1,67 @@
...
+}) => {
+  const { setValue, removeValues } = React.useContext(StoreContext)
+
+  useEffect(() => {
+    const script = document.createElement('script')
+    script.src = 'https://dc.insurely.se/v2/assets/js/dc-bootstrap.js'
+    script.async = true
+    document.body.appendChild(script)
+
+    window.addEventListener('message', (event) => {
```
robinandeer, line 31: We should add a clean up function for this event listener

src/Parsing/parseStoryData.ts
```diff
@@ -581,6 +601,15 @@ const getAction = (containerElement: Element, translate: Translator) => {
     )
   }
 
+  const insurelyIframeNode =
+    containerElement.getElementsByTagName('insurelyIframe')[0]
```
robinandeer, line 605: I don't remember what conventions we follow in Embark but usually kabab case is used in XML, no? `insurely-iframe`.

robinandeer (commented):
Cool stuff looks a lot better integrated than I thought it would!



Have you tested how it works on mobile as well?

I'm not sure if we need to name it `iframe`. That seems like an implementation detail that editor don't really need to care about. Maybe just `InsurelyAction` <= I think it's good to keep the action-suffix convention.

src/Parsing/parseStoryData.ts
dendoesit: I'm not sure we use any case. 

image

dendoesit:
Renamed component to InsurelyAction in order to fit with the rest. Added a cleanup function for the event.

src/Components/Actions/InsurelyAction/InsurelyAction.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  useEffect(() => {
+    const script = document.createElement('script')
+    script.src = 'https://dc.insurely.se/v2/assets/js/dc-bootstrap.js'
+    script.async = true
+    document.body.appendChild(script)
+    const handleData = (event: any) => {
+      const { data } = event
+      if (data.name === 'status') {
+        console.log('Collection status:', data.value)
```
robinandeer, line 33: Would be nice to clean up `console.log` statements before merging!

robinandeer (commented):
Nice update!

I think we also need to expose this new component to the GraphQL schema. Not 💯 on what files you need to change but should be possible to figure out by looking at an example of an existing component!

sampettersson (commented):
Any reason for not replacing the ExternalInsuranceProviderAction with the iframe in your client code? In the apps everything is hooked up to `ExternalInsuranceProviderAction` already and we are not planning on using the iframe :)

src/Components/Actions/InsurelyAction/InsurelyAction.tsx
dendoesit: I left these on purpose so we can see what the iframe logs.
They will be removed in Version 2 after Insurely update the iframe and we connect this to the backend .

dendoesit:
@sampettersson I believe we wanted to test the iframe for a bit before we completely replace the ExternalInsuranceProviderAction on the client code.

robinandeer:
Hmmm, that's an interesting point @sampettersson. Could be just an implementation detail perhaps. Maybe we can find another way to test this feature?

I'm thinking if we want to avoid having to use multiple Embark-flows for web vs. app.

sampettersson:
Aslong as all platforms produce the same result for the action, how it looks and behaves doesn't really matter, that's gonna be very platform dependant, and I think that is fine. Would be nice as you say @robinandeer to avoid creating multiple different flows when its not strictly necessary 😄

dendoesit:
@sampettersson Keep in mind this will be used as a POC . The plan is to have this replace the ExternalInsuranceProviderAction after it has been tested properly. I will shortly create a new flow for the InsurelyAction as well that we will use in web-onboarding in order to do the testing. Either way a new flow will be created in order to have the new comparison flow, design here : https://www.figma.com/file/NvrNcAYRC0Zm2Ic2HWjpvd/Insurely-Iframe ( not final version )

sampettersson:
Aight, that sounds good :)

robinandeer (approved):

gustaveen (commented):
Looks good Dan! 🙌 Just a comment on the width :)

gustaveen (approved):
🙌

PR #721, opened 2021-12-07
Implement `QuoteCartOfferRedirect` component

robinandeer:
## What

Based on `VariantedOfferRedirect`.

Accept a quote cart ID + a list of pre-selected insurance types.

## Why

I think this API is better suited for the new API. It will be harder to save quote IDs to the embark store compared to the existing API. Also insurance types are known ahead of time which hopefully will make it easier for Embark editors to reason about it.

Please come with feedback on this! I will start a new example Angel flow to see how this component will work.


src/Parsing/parseStoryData.ts
```diff
@@ -100,6 +100,38 @@ const parseVariantedOfferRedirect = (containerElement: Element) => {
   return variantedOfferRedirects
 }
 
+/*
+ * <QuoteCartOfferRedirect id="key1" insuranceTypes="key2, key3"></QuoteCartOfferRedirect>
```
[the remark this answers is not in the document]
robinandeer: Yeah, it's just a reminder how it's intended to be used. We have a similar example above so I included it here also.

mehdielket (dismissed):

schema.ts
```diff
@@ -172,6 +172,17 @@ const typeDefs = /* GraphQL */ `
...
 
+    type EmbarkQuoteCartOfferRedirect {
+        component: String!
+        data: EmbarkQuoteCartOfferRedirectData!
+    }
+
+    type EmbarkQuoteCartOfferRedirectData {
+        expression: EmbarkExpression!
+        id: String!
+        insuranceTypes: [String!]!
```
guilhermespopolin, line 183: We use an `Enum` in `giraffe` to represent that info. Is that a way to share it between both places?

guilhermespopolin (commented):
To be honest I didn't understand much 😅 
I believe I'll need some experience working with Embark in order to get a good understanding of its inner work.

src/Parsing/parseStoryData.ts
```diff
@@ -100,6 +100,38 @@ const parseVariantedOfferRedirect = (containerElement: Element) => {
...
+
+const parseQuoteCartOfferRedirect = (containerElement: Element) => {
+  const quoteCartOfferRedirects = Array.from(
+    containerElement.getElementsByTagName('QuoteCartOfferRedirect'),
+  ).map((quoteCartOfferRedirect) => {
+    const whenAttribute = quoteCartOfferRedirect.getAttribute('when')
+    const expression = parseExpression(whenAttribute || '')
+
+    const id = quoteCartOfferRedirect.getAttribute('id')
+    const insuranceTypes =
```
gustaveen, line 115: `insuranceTypes` is the selected ones right? Kind of like naming it like that then `selectedInsuranceTypes` :)

gustaveen (commented):
Looks good! But will be easier to test in a flow of course :) Just had a question on `insuranceTypes`

schema.ts
robinandeer: I think that's some unfortunate naming -- it's not the same as we refer to here. There's no enum for that - I guess since we wanna keep it generic.

src/Parsing/parseStoryData.ts
robinandeer: yeah sure, we can name it like that :)

gustaveen (approved):
🛒

guilhermespopolin (approved):

PR #731, opened 2021-12-10
Support setting variables with a path

robinandeer:
## What

This adds new functionality to enable more flexible variables in `GraphQLAPI` component.

```xml
<Variable key="input.payload[1].firstName"
                 from="firstName"
                 as="string"
/><Variable>
```

Will set the following variable to pass to GraphQL query/mutation:

```json
{
  "input": {
    "payload": [{
      "firstName": "John"
    }]
  }
}
```

- Use lodash set.
- Add test for new functionality.
- Enable ES module interop to make tests work.

## Why

We need this for new Quote Cart API. It will also make it more flexible to set variables.
Needs to be implemented in Android and iOS as well.


hugokallstrom (approved):
LGTM!

sampettersson (approved):
🎉

PR #732, opened 2021-12-10
Update to Web Onboarding NO - English Combo

sampettersson:
Add `ExternalInsuranceProviderAction` to an NO flow, it checks agains the store for `Supports.DataCollectionNorway`, if its not supported it fallbacks to non insurely as previously.


hugokallstrom (approved):
When/how do we add `Supports.DataCollectionNorway` to the store?

sampettersson:
@hugokallstrom I added to the iOS implementation so that it checks for this keyword and just returns `true` when it finds it 😄 I think that works kind of nicely. https://github.com/HedvigInsurance/ugglan/pull/421/files#diff-53dad1c3f396d45418ea55e5ab9b62a754422b90cd933b0c4c9fa39ebbc5f92dR90

sampettersson:
This way we can launch updated stories and the clients will have to tell the story if it supports a specific feature, me and @robinandeer was thinking of using something similar for quoteCart like `Supports.QuoteCart`.

hugokallstrom:
> This way we can launch updated stories and the clients will have to tell the story if it supports a specific feature, me and @robinandeer was thinking of using something similar for quoteCart like `Supports.QuoteCart`.

Cool! 👍

PR #734, opened 2021-12-10
Parse expression fallback to truthy key check

sampettersson:
Allows us to do just `when="Supports.SomethingCool"` instead of `when="Support.SomethingCool == true"` 😄 

Screenshot 2021-12-10 at 13 49 58


robinandeer (approved):
I think this makes sense :)

PR #735, opened 2021-12-10
Add `ConstantVariable` component

robinandeer:
## What

Implement new component to pass constant variables to a GraphQL mutation/query:

```xml
<GraphQLApi next="[[OfferSuccess]]">
  <Mutation>...</Mutation>
  <ConstantVariable key="input.payload[0].data.type"
                    value="SWEDISH_ACCIDENT"
                    as="string"
  >
  </ConstantVariable>
</GraphQLApi>
```

- Update tests to include new component.

## Why

We need this for new Quote Cart API. It also seems like something we want to have the option to do.


gustaveen (approved):
Cool! Just a question, what would `input.payload[0].data.type` resolve to in that example?

robinandeer:
@gustaveen That would resolve to:

```json
{
  "input": {
    "payload": [{
      "data": {
        "type": "SWEDISH_APARTMENT"
      }
    }]
  }
}
```

It's using the same syntax as lodash set :)

src/Components/API/graphql.ts
```diff
@@ -45,13 +46,18 @@ const isSingleVariable = (variable: Variable): variable is SingleVariable => {
...
-export const parseSingleVariable = (variable: SingleVariable, store: Store) => {
-  const value = store[variable.from]
-  if (value === undefined || value === null) {
-    return null
+const isConstantVariable = (
+  variable: Variable,
+): variable is ConstantVariable => {
+  if (variable.__typename === 'EmbarkAPIGraphQLConstantVariable') {
+    return true
   }
```
guilhermespopolin, line 54: [Just a suggestion]
```suggestion
const isConstantVariable = (
  variable: Variable,
): variable is ConstantVariable => variable.__typename === 'EmbarkAPIGraphQLConstantVariable'
```

src/Parsing/parseGraphQLApi.ts
```diff
@@ -26,6 +30,13 @@ export interface MultiActionVariable {
   variables: Variable[]
 }
 
+export interface ConstantVariable {
+  __typename: string
```
guilhermespopolin, line 34: [Question] Shouldn't this be `__typename: ConstantVariable`? 🤔

guilhermespopolin (commented):
Too late review. My bad

src/Parsing/parseGraphQLApi.ts
robinandeer: @guilhermespopolin yeah I thought so at first but just followed the convention based on other types.

hedvigoscar:
I guess I'm a bit late to the party, but couldn't you just inline the constant in the query body to achieve essentially the same result? 🤔

robinandeer:
@hedvigoscar We could before but with quote cart API we build up more complex nested queries using #731 

At least I couldn't find another way around it.

PR #736, opened 2021-12-14
NEW FLOW: Web Onboarding SE - Quote Cart Needer

robinandeer:
## What

A quote cart enabled needer flow in Sweden.

### Changes

- add computed store value for "number of co-insured"
- add new mutation to create quote bundle
- expect quote cart ID to be passed in initial store
- use new `QuoteCartOfferRedirect`
- use new GraphQL variable key syntax (`input.payload[0].firstName`)
- use new `ConstantVariable` component

## Why

- it's duplicated to make it possible to test parallell to the existing flow
- I named the mutations very explicitly so we can track them in Apollo studio (`CreateSwedishNeederApartmentQuoteBundle `)


mehdielket (approved):
Nice!

guilhermespopolin (approved):

PR #738, opened 2021-12-15
 added href target to Message and Footer

dendoesit:
What ? 

- Added target _blank to links inside Message and Footer component

- Upgrading react-markdown to version 7 brings some unwanted typecheck errors so I left it at 6.0.3.
 --> https://github.com/remarkjs/react-markdown/blob/main/changelog.md

image

Why ?

Clicking a link opens in the same tab breaking the onboarding flow.


package.json
```diff
@@ -59,7 +59,7 @@
     "react-dom": "^16.13.0",
     "react-hot-loader": "^4.12.19",
     "react-input-mask": "^2.0.4",
-    "react-markdown": "^6.0.3",
+    "react-markdown": "^7.1.1",
```
robinandeer, line 62: Great if you can update the PR with a reference to breaking changes! And preferably state that we don't need to consider them :)

src/Components/Footer/Footer.tsx
```diff
@@ -41,7 +41,9 @@ export const Footer = (props: FooterProps) => {
   }
   return (
     <TextContainer>
-      <ReactMarkdown>{replacePlaceholders(store, text)}</ReactMarkdown>
+      <ReactMarkdown linkTarget={'_blank'}>
```
robinandeer, line 44:
```suggestion
      <ReactMarkdown linkTarget="_blank">
```

src/Components/Message/Message.tsx
```diff
@@ -36,7 +36,9 @@ export const Message = (props: MessageProps) => {
     <MessageContainer>
       <MessageAnimation>
         <MessageBody isResponse={props.isResponse}>
-          <ReactMarkdown>{replacePlaceholders(store, text)}</ReactMarkdown>
+          <ReactMarkdown linkTarget={'_blank'}>
```
robinandeer, line 39:
```suggestion
          <ReactMarkdown linkTarget="_blank">
```

robinandeer (dismissed):
Nice nice! Just think we should state that we looked into potential breaking changes

We also need to make sure we can make the same upgrade in `web-onboarding`

src/Components/Footer/Footer.tsx
```diff
@@ -41,7 +41,9 @@ export const Footer = (props: FooterProps) => {
   }
   return (
     <TextContainer>
-      <ReactMarkdown>{replacePlaceholders(store, text)}</ReactMarkdown>
+      <ReactMarkdown linkTarget="_blank">
```
gustaveen, line 44: Does this render a link with the attribute `rel="noopener"`? @dendoesit Otherwise we should add it as a security measure when opening links in new tabs

dendoesit: good point. will add https://github.com/rehypejs/rehype-external-links for this as.

robinandeer (approved):
Looks good!

src/Components/Footer/Footer.tsx
dendoesit: took a different approach. @robinandeer helped me build the rel="noopener" in a different component.

PR #739, opened 2021-12-16
GRW-666 / Fix / Correctly check if passage contains redirect.

robinandeer:
## What

Check length of array since empty array `[]` is "truthy" in JS.
Remove unused imports.

> Side note: this didn't cause issues earlier since @sampettersson mistyped `variantedOfferRedirect` in his PR => which will always return `undefined` which is "falsy". When I corrected it to `variantedOfferRedirects` it returns an empty array ("truthy") and this issue emerged 😅 

## Why

Fix issue causing back button to disappear.


dendoesit (approved):
great 🎉

gustaveen (approved):
Oh, good find Robin! 👌

PR #740, opened 2021-12-21
Update to Norwegian Switcher

hugokallstrom:
This is a new story which uses text keys from lokalise, otherwise it is a copy from "Web Onboarding NO - English Contents". We will extend this story with other questions similar to how the Swedish stories are structured. Another PR will contain "Norwegian Needer" - with no questions related to switching, also similar to the Swedish stories.


robinandeer (approved):
Looks good - we can fix Lokalise for keywords in another PR!

PR #747, opened 2022-01-04
NEW: Web Onboarding NO - Quote Cart Norwegian Combo

robinandeer:
Uses `SE_WEB_SIGNUP_OFFER_UW_LIMIT_ERROR_MESSAGE_1` since it has norwegian translations but should use a new key.


mehdielket (approved):
🚀

robinandeer:
Thanks for the approve @mehdielket but I think we are closing this to build on #740 instead so we get a localised flow!

robinandeer:
Replaces by #749

PR #750, opened 2022-01-11
fix(GRW-716): parse story keyworkds in an immutable way

guilhermespopolin:
## What

Back button wasn't being translated properly.

Screenshot 2022-01-11 at 11 07 48

## Why

Story keywords were being udated by reference. That was making subsequent requests for story data return, as keyword values, the result of the translation taken in place during the initial load.

```
// story data
{
   ...,
   keywords: {
      ...,
      backButton: '{SE_WEB_SIGNUP_BACK_BUTTON}'
   }
}
```

_Page Load / First Translation result (locale sv_SE)_

```
// story data
{
   ...,
   keywords: {
      ...,
      backButton: 'tillbacka'
   }
}
```

_Subsequent requests result (locale en_SE)_

We use [resolvePlaceholders](https://github.com/HedvigInsurance/embark/blob/master/src/Parsing/parseStoryData.ts#L1071) dependency in order to handle translation placeholders (`{SE_WEB_SIGNUP_BACK_BUTTON}`) and make the translations. By calling it with story data described bellow the result will be the same, even though we changed the locale while making the request.

```
// story data
{
   ...,
   keywords: {
      ...,
      backButton: 'tillbacka'
   }
}
```

[Ticket](https://hedvig.atlassian.net/browse/GRW-716)


sampettersson (approved):
Super weird bug! Good find!

mehdielket (approved):
Nice catch and great job with the investigation 👏

gustaveen (approved):
Great digging @guilhermespopolin! 🙌

PR #755, opened 2022-01-20
Update to Web Onboarding SE - Needer

robinandeer:
Fix a small issue when users go back and forth in the flows.

Issue comes from the fact that we override "apartmentType" in the "are you student" question. This PR instead stores the BRF/RENT value in a separate "rentOrBrfType" field that doesn't change later in the flow.

This value is then checked in the "student eligibility" passage instead of "apartmentType".


styrelius (approved):
👍

PR #759, opened 2022-01-24
fix(GRW-732): proper render privacy policy link

guilhermespopolin:
# What

- Updated `SE_WEB_SIGNUP_KEYWORD_INSURELY_CONFIRM_PRIVACY_POLICY` key so it defines _Privacy Policy_ link using markdown;
Screenshot 2022-01-24 at 12 26 26

- Updates `ConfirmCollectionStep` component so it parses `SE_WEB_SIGNUP_KEYWORD_INSURELY_CONFIRM_PRIVACY_POLICY` key using `react-markdown`;

# Why

image

[Ticket](https://hedvig.atlassian.net/browse/GRW-732)


robinandeer (commented):
Nice work!

I'm just thinking how we will deploy this.

1. update translations (add new key?)
2. update embark flow to add new translations + deploy
3. update web-onboarding with new embark-dependency

So then we would need a new story keyword, right?

robinandeer:
Hmm yeah I wasn't really sure how it would work -- but if we won't break the existing setup/flow in the mean time than we can just go ahead :)

robinandeer (approved):
Nicely done!

PR #761, opened 2022-01-24
Update to moving-flow-DK

sampettersson:
* Fix isYouth being passed instead of isStudent
* Add autocomplete address
* Set max student apartment size to 100 sqm


hugokallstrom (approved):
LGTM

PR #763, opened 2022-01-25
Update to Web Onboarding SE - Switcher Without Accident

StylianosGakis:
Since this is the switcher flow, we know that isMover should always be true.
This change is made since as it is now, on both iOS and Android since the key is not present both when statements are not triggering and no text is shown to the member.

The change is going from
```
<Message>
    <When expression="isMover == 'false'">
        {SE_WEB_SIGNUP_NEEDER_HOME_TYPE_MESSAGE_1}
    </When>
    <When expression="isMover == 'true'">
        {SE_WEB_SIGNUP_HOME_TYPE_MESSAGE_1_MOVER}
    </When>
</Message>
```
to
```
<Message>
  {SE_WEB_SIGNUP_HOME_TYPE_MESSAGE_1_MOVER}
</Message>
```


sampettersson (approved):

hugokallstrom (approved):
LGTM

PR #774, opened 2022-01-27
Web Onboarding NO - Contents

hugokallstrom:
This story will replace "Web Onboarding NO - English Contents" and "Web Onboarding NO - Norwegian Contents". All hard coded strings have been replaced with string keys, which means the story can be used for locales.


robinandeer (approved):
Awesome! Have we removed all trackable properties?

hugokallstrom:
> Awesome! Have we removed all trackable properties?

Good catch. Added same trackable properties as "English Contents".

robinandeer (approved):

PR #793, opened 2022-02-08
fix: whitelist files that should be included into embark package

guilhermespopolin:
## What

Whitelist that should be included into `embark` package by using `package.json` `files` field;

## Why

We are currently including all the files which might end up being a security issue:
all-files

There's a `.npmignore` file at the root of `embark` package that were suppose to avoid this issue as it's declaring that everything rather then `dist/` should be ignore while composing the package. But it seems that, [when placed in the root of the project](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#:~:text=At%20the%20root%20of%20your%20package%20it%20will%20not%20override%20the%20%22files%22%20field%2C%20but%20in%20subdirectories%20it%20will.), that file will not override `package.json` `files` field [which is `[*]` by default (include everything)](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#:~:text=Omitting%20the%20field%20will%20make%20it%20default%20to%20%5B%22*%22%5D%2C%20which%20means%20it%20will%20include%20all%20files.).

One can use `npm pack` or `npx npm-packlist` in order to check the files that will be included in the `embark` package. Below you can see the results of `npm pack` command before and after this new setup.

**Before**
before

**After**
after

## How to test

Here's how I've tested:

1. While on this feature branch, generate a new package by running `npm pack` (you can take this opportunity to also double check the files that were included into the package since they are logged during the execution of this command);
2. Go to `web-onboarding` and reinstall `embark` using the package generated in the previous step: `yarn add <path-to-package.tgz>`. At the end of this command you can check that `embark` dependencies under `node_modules/@hedviginsurance` has only the files listed on step 1;
3. Run `web-onboarding` locally with `yarn watch`. You should be able to complete _Embark_ flows as before;


robinandeer:
@guilhermespopolin Let's maybe just team up on this and merge/compete it together :)

PR #794, opened 2022-02-11
Set correct type for quote cart stories

hugokallstrom:


StylianosGakis (approved):

StylianosGakis (approved):
For knowledge sharing purposes, how is this the used to affect how the apps handle it? Is there some special matching of these names, or? Why does this "type" exist?

hugokallstrom:
> For knowledge sharing purposes, how is this the used to affect how the apps handle it? Is there some special matching of these names, or? Why does this "type" exist?

Yes we are filtering stories in the clients on type. So we would ideally want to have a feature flag and switch from "APP_ONBOARDING" to "APP_ONBOARDING_QUOTE_CART" when quote cart is implemented.

PR #795, opened 2022-02-11
Remove app type from NO stories with hard coded text

hugokallstrom:


hugokallstrom:
> Again for knowledge sharing, I am not 100% sure what this metadata is used for and why it wouldn't be needed anymore? Ready to learn 🤓

These stories are replaced by "Web onboarding NO - Combo" and "Web onboarding NO - Contents", so we should not display them in the app anymore. Thats why the metadata is removed. We can remove the stories themselves when web has tested that everything works OK on their end.

PR #813, opened 2022-02-28
Add separate "from" key in `MultiActionVariable`

robinandeer:
Use "key" as default value to maintain backwards compatibility.


mehdielket (approved):
🚀

PR #824, opened 2022-03-17
fix(GRW-915): handle chained redirect passages

guilhermespopolin:
## The problem

Consider the following _Embark_ flow snippet where both _youthEligibilityCheck_ and _OfferSelector_ are _redirect passages_:

Screenshot 2022-03-17 at 20 16 20
youth-eligility
offer-selector

On the scenario where we transition from _youthEligibilityCheck_ to _OfferSelector_ the following occurs:

https://user-images.githubusercontent.com/19200662/158881711-6c7ac244-7142-49c9-a9f9-835af0fb54d6.mov

That's happening because the implementation doesn't take into account that a passage we're redirecting to might be another _redirect passage_ (chained redirect). On this particular case, we end up rendering _OfferSelector_ which doesn't have any messages - therefore the "This passage has no messages" message.

## The solution

* Refactoring: move `useGoto` functionality into `useEmbark` hook, moving away from a _change passage_ feature that's based on _effects_ which are harder to understand (IMHO)
* Make usage of _recursion_ to handle internal redirects: when we face an internal redirect passage (passage that makes usage of <Redirect />) we:
    * render that passage as usual when the criteria for redirection is not matched;
    * call  `goTo` function with the name of the passage which criteria for redirection is matched;


mehdielket (approved):
🚀

PR #825, opened 2022-03-22
fix(GRW-915): handle chained redirect passages

guilhermespopolin:
## The problem

Consider the following _Embark_ flow snippet where both _youthEligibilityCheck_ and _OfferSelector_ are _redirect passages_:

Screenshot 2022-03-17 at 20 16 20
youth-eligility
offer-selector

On the scenario where we transition from _youthEligibilityCheck_ to _OfferSelector_ the following occurs:

https://user-images.githubusercontent.com/19200662/158881711-6c7ac244-7142-49c9-a9f9-835af0fb54d6.mov

That's happening because the implementation doesn't take into account that a passage we're redirecting to might be another _redirect passage_ (chained redirect). On this particular case, we end up rendering _OfferSelector_ which doesn't have any messages - therefore the "This passage has no messages" message.

## The solution

* Updates `useGoto` so it proper evaluate the next passage when we’re dealing with a _internal redirect passage_;
* Added minor code style changes

## Ticket
[GRW-915](https://hedvig.atlassian.net/browse/GRW-915)

## Relates to
#824 - This one was an unsuccessful attempt to tackle the problem in a different way. Even though it was merged it was reverted in `master`


src/Utils/ExpressionsUtil/useGoTo.ts
```diff
@@ -12,38 +12,34 @@ import { passes } from './expression'
...
-      const targetPassage = newPassage ? newPassage.id : data.startPassage
+      )
+      const targetPassageId = newPassage ? newPassage.id : data.startPassage
 
       if (newPassage.redirects.length > 0) {
-        const passableExpressions = newPassage.redirects.filter(
-          (expression: any) => {
-            return passes(store, expression)
-          },
+        const passableExpressions = newPassage.redirects.find(
```
robinandeer, line 30: Rename this variable since it's not an array anymore

```suggestion
        const passingExpression = newPassage.redirects.find(
```

robinandeer (dismissed):
That's a very React-y way to do recursion 😅 but it should work I suppose

And nice clean-up

guilhermespopolin:
> That's a very React-y way to do recursion 😅 but it should work I suppose
> 
> And nice clean-up

Indeed. I wasn't my first try. I tried to reach the same result with a different approach on #824 but unfortunately no success 😢

robinandeer (approved):
Re-approve!

PR #833, opened 2022-03-30
feat(GRW-936): select action with single option

guilhermespopolin:
## What

Updates _SelectAction_/_SelectOption_ actions so we can proper cover the scenario where this only one option available: displays a button instead of a card.

## Why

We have passages in _NO Pentagon Accident_ flow that just inform something to users and require from them that they move forward to the next passage, acknowledging that they are aware of whatever we are telling. E.g:
accident

https://user-images.githubusercontent.com/19200662/161041524-6009c9a0-fe04-4485-95c0-553f83de34ff.mov

## Relates with

#834 

### [Ticket](https://hedvig.atlassian.net/browse/GRW-936)


src/Components/Actions/Action.tsx
```diff
@@ -161,5 +163,14 @@ export const Action = (props: ActionProps) => {
     )
   }
 
+  if (props.action.component === 'MoveForwardAction') {
+    return (
+      <MoveForwardAction
```
guilhermespopolin, line 168: I know that `<MoveForwardAction />` is just an additional layer that, in the end, it's just using `ContinueButton` component. However it 'feels weird' don't have it and just use `ContinueButton` directly here, right?! I'm not sure. Any opinions?

mehdielket (dismissed):
This looks great to me. Two things
* We need to inform & get the opinion of the app team about this so that they implement this tag.
* We should somehow add this in the Embark doc.

robinandeer (dismissed):
Looks good!

Remember that we are not doing schema stitching anymore so we need to update the schema in Giraffe for this new component as well 😌

hugokallstrom:
We support this in the apps by using a <SelectAction> with one <Option>. For that case we are showing a centered "primary" button. Otherwise we show a grid. 

Would that solution be messy on web?

guilhermespopolin:
> We support this in the apps by using a `<SelectAction>` with one `<Option>`. For that case we are showing a centered "primary" button. Otherwise we show a grid.
> 
> Would that solution be messy on web?

I'm investigating but upfront it seems like a better idea.

src/Components/Actions/SelectAction/SelectOption.tsx
```diff
@@ -82,26 +81,37 @@ const TooltipWrapper = styled.div`
...
 
-export const SelectOption = (props: SelectOptionProps) => {
+export const SelectOption = ({
+  label,
+  onClick,
+  isSingleOption = false,
+  badge,
+  tooltip,
+}: SelectOptionProps) => {
+  if (isSingleOption) {
```
guilhermespopolin, line 100: This is the most important change: we display a button in case only one option is provided.

mehdielket (approved):
Nice that we are now in sync with the app team and how they do things.
 
 One small nit: The "continue" shows up as an answer to the message. Ideally we probably don't want any answer. Maybe this is a large effort, and if so, then it's not worth it I think

guilhermespopolin:
> Nice that we are now in sync with the app team and how they do things.
> 
> One small nit: The "continue" shows up as an answer to the message. Ideally we probably don't want any answer. Maybe this is a large effort, and if so, then it's not worth it I think

Yeah, I thought it would be a good thing to have. Additionally, with this approach no changes on flow will be required.

I saw that you've mentioned in the ticket but I failed to communicate the decision of nothing adding it here, my bad.
But you could guess the reason: it seems to be a large effort

mehdielket (approved):
🚀

PR #834, opened 2022-03-30
feat(GRW-936): makes usage of MoveForwardAction

guilhermespopolin:
## What

Updates `onboarding-NO` flow so it makes usage of `<MoveForwardAction />`

## Why

Part of #833 

## Relates with

#833

### [Ticket](https://hedvig.atlassian.net/browse/GRW-936)


mehdielket:
I guess this one is to be closed right?

guilhermespopolin:
> I guess this one is to be closed right?

Yes. I'm just holding it until #833 is not merged

PR #835, opened 2022-03-30
styles(angel-preview): remove scroll bar

guilhermespopolin:
## What

Removes vertical scroll bar from _angel preview_

## Why

Its presence can be annoying sometimes while testing flows locally

## Before

before-scroll

## After

after-scroll


gustaveen:
Hm, I assume te scrollbars only show up from the Angel preview. Since when you look at embark in the context of the onboarding the scrollbars don't show up. It would be nice to solve it the same way for both cases or WDYT :) 

We calculate the height for the `ChatContainer` if `hasHeader` is true. 
https://github.com/HedvigInsurance/embark/blob/master/src/Components/Passage.tsx#L33
Shouldn't we just pass `hasHeader={true}` for `<Passage>` and we would have the same behaviour :)

gustaveen (commented):
Hm, I assume the scrollbars only show up in the Angel preview. Since when you look at embark in the context of the onboarding the scrollbars don't show up. It would be nice to solve it the same way for both cases or WDYT :) 

We calculate the height for the `ChatContainer` if `hasHeader` is true. 
https://github.com/HedvigInsurance/embark/blob/master/src/Components/Passage.tsx#L33
Shouldn't we just pass `hasHeader={true}` for `<Passage>` and we would have the same behaviour :)

guilhermespopolin:
> Hm, I assume te scrollbars only show up in the Angel preview. Since when you look at embark in the context of the onboarding the scrollbars don't show up. It would be nice to solve it the same way for both cases or WDYT :)
> 
> We calculate the height for the `ChatContainer` if `hasHeader` is true. https://github.com/HedvigInsurance/embark/blob/master/src/Components/Passage.tsx#L33 Shouldn't we just pass `hasHeader={true}` for `<Passage>` and we would have the same behaviour :)

Awesome 🎉 
Soo much better. Thanks @gustaveen

gustaveen:
Nemas problems! :relaxed: Awesome that you took the time to fix this @guilhermespopolin, it's been bothering me many times! 😆

gustaveen (approved):
:star:

PR #889, opened 2022-04-22
Add GH workflows for CI and NPM publish

robinandeer:
The CI workflow will run for each PR and pushes to master.

The publish workflow will run on pushes to master.

> Only publishes if the version number in `package.json` differs from the latest on NPM

Questions:

1. Therefore I feel like we don't need to involve either an approval step or tags, what do you say?
2. Should we keep the Slack notification?

FYI: we need to add a new secret for the `NPM_TOKEN` to make this work:

https://github.com/HedvigInsurance/brand/settings/secrets/actions/new


.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,36 @@
...
+
+on:
+  push:
+    branches:
+      - master
+
+  workflow_dispatch:
+
+jobs:
+  test:
```
simonauner, line 11: nit:
```suggestion
  publish:
```

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,36 @@
+name: Publish to NPM
```
simonauner, line 1: Shouldn't this build either contain the steps from the file above (and removing `on:  push:    branches:      - master`) or depend on that job? It looks like these two will run independently of each-other and will thereby publish to NPM even if testing fails.

(I know it shouldn't happen since we have tests for any PR that merges to master, but still)

Maybe just copy paste or create a reusable workflow? https://docs.github.com/en/actions/using-workflows/reusing-workflows#example-caller-workflow

simonauner (commented):
Nice work! It's great that we get this on GH 👍 
Just a question about testing and typechecking when publishing

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,36 @@
...
+
+on:
+  push:
+    branches:
+      - master
+
+  workflow_dispatch:
+
```
sampettersson, line 9: FYI workflow dispatches can take input, I usually let it take in the version the user wants to publish and then update all necessary files push a commit, tag and make a GH Release all in the workflow. Then the only thing the user needs to do to publish is to run the workflow 😄

.github/workflows/publish-npm.yml
robinandeer: Yeah my idea was that it would already be tested in a PR but maybe it doesn't hurt to include here as well.

.github/workflows/publish-npm.yml
robinandeer: Interesting.

Yeah perhaps we shouldn't worry too much about auto-publishing to NPM. Could be better to just run this workflow on demand.

simonauner (dismissed):
⭐ 

Let's just remember to update the documentation on how to release going forward since we now do it in GH instead of Codefresh 🙌

simonauner (approved):

PR #890, opened 2022-04-25
GRW-998 / Feature / Add Car registration number mask

simonauner:
This PR adds a mask for Car registration number.

~~I would like it to transform to uppercase when the user inputs, but I couldn't find that we do any such transform today?~~

We now handle to uppercase via CSS for the input component. Perhaps we should look into something like https://github.com/sanniassin/react-input-mask/issues/169 because now when the user enter `abc123` it displays as `ABC 123` in the input field, but the response chat bubble still says `abc 123`. But perhaps that is for a future PR.

https://user-images.githubusercontent.com/1343979/165068093-9c4cd113-19ef-4f08-83d2-6537862f608d.mov


src/Components/Actions/masking.test.tsx
```diff
@@ -50,6 +50,22 @@ describe('isValid', () => {
     expect(isValid('BirthDateReverse', invalidBirthDate)).toBe(false)
   })
 
+  describe('validates CarRegistrationNumber mask', () => {
+    it.each(['ABC123', 'ABC12A'])(
+      'returns true for valid registration number %p',
+      (valid) => {
+        expect(isValid('CarRegistrationNumber', valid)).toBe(true)
```
sampettersson, line 57: Should we maybe call it `SwedishCarRegistrationNumber`? Or do you think this mask applies to other countries?

sampettersson (dismissed):
Nice! Should probably do this in the apps too :) Just a thought regarding naming.

src/Components/Actions/masking.test.tsx
simonauner: Yes, very valid point :D

simonauner: Fixed!

simonauner:
Renamed to `SwedishCarRegistrationNumber` and added a space between letters and numbers: `ABC 123`

sampettersson (dismissed):
👌

README.md
```diff
@@ -18,7 +18,7 @@ Steps to deploy:
 - Wait for Codefresh to build the docker container. It will be deployed to dev automatically, you can see when it's deployed in the [#platform-deploy](https://hedviginsurance.slack.com/archives/C01PGTP0VK4) Slack channel.
 - Use the deployment helper script in prod-env to deploy to prod (see separate instructions in that repository). You can follow the progress in hte [#platform-deploy](https://hedviginsurance.slack.com/archives/C01PGTP0VK4) Slack channel
 
-To develop: use `yarn start`
+To develop: use `yarn "watch`
```
robinandeer, line 21:
```suggestion
To develop: use `yarn watch`
```

robinandeer (approved):

README.md
simonauner: 🦅 👁️ !
Fixed!

robinandeer (approved):
Ser bra ut :)

simonauner:
Let's try it :D

PR #893, opened 2022-04-25
Add support for hidden links

sampettersson:
When calculating the maxDepth of a story the calculation in both apps and web most often times include a last "error" passage, making it look a bit off. This PR adds a way to say that a link shouldn't be used for length calculation, its thus "hidden".

How things looks in the apps currently:

Also took the chance to fix loops in stories making the progress bar break on web, we had previously fixed that in native [clients](https://github.com/HedvigInsurance/ugglan/blob/81e7096a40f609546ff12f3b2b5b47b684d8e155/Projects/Embark/Sources/EmbarkState.swift#L197).


robinandeer (commented):
Looks cool!

Feel like adding a couple tests? ;)

robinandeer (approved):
Let's go!

robinandeer (approved):

PR #907, opened 2022-05-02
Update to SE-cross-sell-car

hugokallstrom:
Export keys to lokalise. Re-use as many keys from the web onboarding flow as possible.


sampettersson (approved):
⭐

PR #916, opened 2022-05-05
GRW-1061 - Update to Web Onboarding SE - Switcher Without Accident student eligibility rules

mehdielket:
New student eligibility rules:

- Home size up to 250m2
- Household size up to 5


gustaveen (approved):

robinandeer:
@mehdielket We don't need to keep updating these that don't support quote cart I think

mehdielket:
> @mehdielket We don't need to keep updating these that don't support quote cart I think

We are still using these (and the old offer page) for some external partners.

PR #924, opened 2022-05-09
Cannot compare errors with null for web

simonauner:
Even though the request to create a quote bundle succeeds we get redirected to the error page.

Turns out we cannot compare errors with `null` for web, and thus we must instead store typename of response and use that to determine whether to redirect to offer page or error page after bundle is created.


robinandeer (approved):

sampettersson:
Should we maybe fix null checking for web? We have this in both native clients and its a thing that comes up now and then causing bugs and differentiation between platforms 😬

sampettersson (approved):

robinandeer:
@sampettersson yes. But let's make this flow work first 😉

PR #925, opened 2022-05-10
Update to Web Onboarding SE - Quote Cart Switcher

hugokallstrom:
Add 
```
<When expression="isMover == 'null'">
    {SE_WEB_SIGNUP_HOME_TYPE_MESSAGE_1_MOVER}
</When>
```

For when we dont have the `isMover` value in the store.


StylianosGakis (approved):
So it's basically defaulting to what would happen if `isMover` was "true" when it's missing right?

robinandeer (commented):
Do you know if this works in the web implementation? I guess so...

hugokallstrom:
> So it's basically defaulting to what would happen if `isMover` was "true" when it's missing right?

Yes!

> Do you know if this works in the web implementation? I guess so...

Dont know. We never set this `isMover` value in the apps though, so I assumed it was set in web

PR #926, opened 2022-05-10
feat(GRW-864): added a flow that supports house acquisition

guilhermespopolin:
## What

Added a flow (`onboarding-NOv2`) that supports house acquisition.

Gonna test it on stage before overriding `onboarding-NO` - that should be the single flow for NO.

## Ticket
[GRW-864](https://hedvig.atlassian.net/browse/GRW-864)


robinandeer (approved):
Let's go!

PR #929, opened 2022-05-10
Allow disabling translation cache

robinandeer:
Introduce an environment variable to disable translation cache for a given deployment.

@sampettersson @hugokallstrom I was thinking that we can disable the cache in staging -- do you think that's a good idea? I think it would help the content editors.

Do you know where I would add such an environment variable?


sampettersson:
I think @oscarnylander can chime in here, don't remember exactly why we added the cache.

PR #930, opened 2022-05-10
fix(onboarding-NOv2): updated coInsured and livinSpace fields

guilhermespopolin:
## What

`numberOfCoInsured` rather than `coInsured` and `squareMeters` rather than `livingSpace`


mehdielket (approved):
🚀

PR #937, opened 2022-05-13
fix(onboarding-NOv2): make usage of hidden link to improv perf

guilhermespopolin:
See title


sampettersson (approved):
Glad the hidden functionality is coming in handy 🥳

PR #947, opened 2022-05-23
Add new method `transformInput` to masking

simonauner:
## What:

Adds `transformInput` method and call that in `TextAction`'s `onChange` handler

## Why

Previously, the input field was styled to uppercase for `SwedishCarRegistrationNumber` but then when the user submits the input the chat reply bubble shows in lowercase since the actual text was (maybe) input like that. It's better to just transform the input to uppercase as well. This makes it look better in later stages of the flow as well (Offer page in web-onboarding, for example).

https://user-images.githubusercontent.com/1343979/169753105-5f429d50-f004-4a2d-a504-1db4833276dd.mov

I added a new method `transformInput` to solve this, and then call that in a new `onChange` handler:

```jsx
  const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
    setTextValue(transformInput(e.target.value, props.mask))
  }
```

We do sort of the same thing when initializing `setTextValue` hook:

```jsx
  const [textValue, setTextValue] = React.useState(
    mapMaskedValue(store[props.storeKey], props.mask) || '',
  )
```

So I could potentially call it in sort of the same way in the `onChange` handler:

```jsx
  const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
    setTextValue(mapMaskedValue(e.target.value, props.mask))
  }
```

But I'm not sure how that affects `BirthDateReverse` in `mapMaskedValue`:

```jsx
export const mapMaskedValue = (value: string, m?: MaskType): string => {
  if (m === 'BirthDateReverse') {
    try {
      return format(parse(value, 'yyyy-MM-dd', 0), 'dd-MM-yyyy')
    } catch {
      return value
    }
  }

  return value
}
```

Do you have any input on which solution to prefer? Either keep it as it is in this PR or to use `mapMaskedValue`. We could have a call and discuss the options 🤙


src/Components/Actions/TextAction/TextAction.tsx
```diff
@@ -105,6 +106,10 @@ export const TextAction: React.FunctionComponent<TextActionProps> = (props) => {
     }
   }
 
+  const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
```
robinandeer, line 109: nit: avoid single character variable names => `event` :)

src/Components/Actions/masking.tsx
```diff
@@ -125,6 +125,14 @@ export const mapMaskedValue = (value: string, m?: MaskType): string => {
   return value
 }
 
+export const transformInput = (value: string, m?: MaskType): string => {
```
robinandeer, line 128: nit: avoid single char variable names => `mask` / `maskType`

robinandeer (approved):
Nice! Yeah this is probably better way to handle it

Would be nice to co-locate all the logic related to "SwedishCarRegistrationNumber" and other mask types but I'm happy with this :)

simonauner:
Nice, thanks @robinandeer . Fixed your comments now

robinandeer (approved):
Looks good! Nice work

PR #950, opened 2022-05-23
GRW-1132 / Add proper null-check for expressions

simonauner:
## What

This PR adds a conversion of `"null"` strings to `null` value when computing comparisons.

## Why

Currently we cannot create expressions like

```jsx
<Redirect to="[[Mileage]]"
          when="carMakeAndModel != null"
>
</Redirect>
```

for Embark in web, since the right-hand side of the expression will be turned into a string `"null"`. Therefore, when the value of  `carMakeAndModel` is actually `null`, we will compare `null` to `"null"` which is never true. So for the expression above, `carMakeAndModel` is always different from `"null"`, and user always gets redirected to `[[Mileage]]`, even when `carMakeAndModel` is `null`.

---

Since the Store needs to hold a nullable value for the tests, I had to add `| null` to some type definitions. I don't think it fails anything?

---

This PR is a consequence of this PR https://github.com/HedvigInsurance/embark/pull/949 not working since I copied the null check from apps 😇 

[GRW-1132](https://hedvig.atlassian.net/browse/GRW-1132)


sampettersson (approved):
🎉

gustaveen (approved):
Great! 🙌

src/externalRedirect.ts
```diff
@@ -3,7 +3,10 @@ import * as React from 'react'
 export interface TExternalRedirectContext {
   Offer: (ids: string[]) => void
   VariantedOffer: (allIds: string[], selectedIds: string[]) => void
-  QuoteCartOfferRedirect: (id: string, selectedInsuranceTypes: string[]) => void
+  QuoteCartOfferRedirect: (
+    id: string | null,
```
robinandeer, line 7: Is this necessary?

I worry about introducing these changes. What is the case where `id` is `null`? Or does it mean that Embark can't guarantee that it isn't null so we need to handle that in `web-onboarding?

robinandeer (commented):
Looks good! Just had one question about the updated types

src/externalRedirect.ts
simonauner: Since values in the Store can be `null` (the server can respond with `null` values), and the `id` is a Store value, it can be `null`. This change was needed since the Store is now being defined as being capable of storing `null` values (https://github.com/HedvigInsurance/embark/pull/950/files#diff-8d9fb16b16dd22e64269f60933903ec159d7bacbae5ed4cd6459135f0a29af76R5), which I guess it _actually_ did before as well in runtime, it's just that no-one had typed it that way.

Does that make sense?

robinandeer: Right, yes it does.

What I wonder is where this should be handled? In Embark or Web Onboarding?
Seems like this method doesn't really make sense now - but if Web Onboarding is better capable of handling the error I guess this is correct!

simonauner: The typing didn't stop the value from being `null` before though, and I'm not sure where it was handled before 🤔

simonauner: I guess that it should be some error handling in Embark for this.

guilhermespopolin (commented):
I've run a typecheck (cmd+shift+p -> Run Task -> tsc: watch - tsconfig.json) and unfortunately it seems that change "breaks" the types on a bunch of places. The issue is that there're some types which declare optional fields and they rely on values provided by the `store` object, leading to a `'string | null' is not assignable to type 'string | undefined'.` error 😢 

image

simonauner:
> I've run a typecheck (cmd+shift+p -> Run Task -> tsc: watch - tsconfig.json) and unfortunately it seems that change "breaks" the types on a bunch of places. The issue is that there're some types which declare optional fields and they rely on values provided by the `store` object, leading to a `'string | null' is not assignable to type 'string | undefined'.` error 😢
> 

Ok, thanks for looking into this @guilhermespopolin ! I'll have a look and see what we can do. How do we solve this when some values from BE can be `null`, but we trust that others aren't? 🤔

guilhermespopolin:
> > I've run a typecheck (cmd+shift+p -> Run Task -> tsc: watch - tsconfig.json) and unfortunately it seems that change "breaks" the types on a bunch of places. The issue is that there're some types which declare optional fields and they rely on values provided by the `store` object, leading to a `'string | null' is not assignable to type 'string | undefined'.` error 😢
> 
> Ok, thanks for looking into this @guilhermespopolin ! I'll have a look and see what we can do. How do we solve this when some values from BE can be `null`, but we trust that others aren't? 🤔

If I'm not wrong, that's a case for [Non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) (!). I don't like it so much but I guess it's a feature that, when used with caution, can help, like `!important` on CSS.

simonauner:
> > > I've run a typecheck (cmd+shift+p -> Run Task -> tsc: watch - tsconfig.json) and unfortunately it seems that change "breaks" the types on a bunch of places. The issue is that there're some types which declare optional fields and they rely on values provided by the `store` object, leading to a `'string | null' is not assignable to type 'string | undefined'.` error 😢
> > 
> > 
> > Ok, thanks for looking into this @guilhermespopolin ! I'll have a look and see what we can do. How do we solve this when some values from BE can be `null`, but we trust that others aren't? 🤔
> 
> If I'm not wrong, that's a case for [Non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) (!). I don't like it so much but I guess it's a feature that, when used with caution, can help, like `!important` on CSS.

I agree, I feel dirty when using it 🙃 I'll give it a try and push an update to this PR soon!

simonauner:
Added non-null assertion operator in all places tsc complained now 🤢

robinandeer (approved):
Yikes 😱 but if this better represents which assertions we do it's a step in the right direction

simonauner:
> Yikes 😱 but if this better represents which assertions we do it's a step in the right direction

We do not make any assertions at all, we make _assumptions_ 😏 about the values not being null

PR #951, opened 2022-05-23
Remove hidden from links properly

sampettersson:
The regex didn't remove the `hidden:` from the link correctly 😬


robinandeer (approved):
Looks even simpler now

PR #962, opened 2022-05-25
Add a section in CHANGELOG when we publish new NPM versions in Github…

simonauner:
## WHAT 

Add a section in CHANGELOG when we publish new NPM versions in Github Actions

## WHY

I feel it's difficult to trace what each new NPM version contains. We can either create Github releases (there are plenty of actions for this), or we can do something simple like using a CHANGELOG file.

My idea is that you run the Github Action, then checkout the PR branch and edit the CHANGELOG file with some useful content.


robinandeer (approved):
I like this even if it creates some additional steps when releasing a new version.

Perhaps we can suggest that you paste a compare link between the last release and this so you can at least look at the diff?

simonauner:
> I like this even if it creates some additional steps when releasing a new version.
> 
> Perhaps we can suggest that you paste a compare link between the last release and this so you can at least look at the diff?

Yeah, that's a good idea. I thought about adding the git log into the CHANGELOG automatically but since this is also a server repository there are many commits that do not affect the NPM package.

PR #964, opened 2022-05-25
Update to SE-onboarding-car

simonauner:
## What

Change so that when re-entering your registration number it will be stored as a new variable.

Add GraphQL query that tries to find car based on the re-entered registration number.

## Why

Because if it uses the same variable as the first "enter registration number" step, the text in the chat bubble will replace `{registrationNumber}` with whatever you re-enter, which looks strange.

image


dendoesit (approved):
looks good.

PR #967, opened 2022-05-27
fix(house-dk): remove water leakage device from mutations

guilhermespopolin:
## What
Remove `waterLeakageDetector` field from House-DK mutations

## Why
There no such thing for House-DK


dendoesit (approved):
👍

PR #972, opened 2022-06-01
fix(GRW-1166): youth eligibility check takes choseen promises into account

guilhermespopolin:
## What
This updates the flow so it takes insurance choices into account while determining if one is eligible for youth insurance or not. Since we don't offer House Youth insurance, if you have not chosen Home Contents at the beginning of the flow them you're automatically not eligible for Youth insurance.

## Ticket
[GRW-1166](https://hedvig.atlassian.net/browse/GRW-1166)


robinandeer (changes requested):
Hmm this isn't what we discussed with @mehdielket tho...

I think we should change so that if `isHouse == true` then we set `isYouth = false`.

guilhermespopolin:
> Hmm this isn't what we discussed with @mehdielket tho...
> 
> I think we should change so that if `isHouse == true` then we set `isYouth = false`.

You're totally right. My bad

robinandeer (approved):

PR #973, opened 2022-06-01
Remove extra chat bubble in NO House.

dendoesit:


robinandeer (approved):

guilhermespopolin (changes requested):
We also need to update living space passage as it's described in the ticket: don't show the second message bubble if uses has selected house insurance. I guess it would be something like this:
```
<Message>
  <When expression="isHouse == false">
    {NO_SIGNUP_SIZE_MESSAGE_2}
  </When>
</Message>
```

PR #982, opened 2022-06-13
fix(GRW-1201): onboarding no updates

guilhermespopolin:

PR #986, opened 2022-06-14
revert: DOMParser

guilhermespopolin:
## What
Reverts the usage of DOMParser since it's breaking the Embark server. I thought that making it globally available would solve the issue but it still happening for a reason that I couldn't understand why. So until I figure that out my intention is to unblock staging.

## Relates to
#982 - Added DOMParser to support a handling `<When>` tags as children of `<Title>` (`<GroupedResponse>`)


simonauner (approved):
Too bad the other solution didn't work. It's a good idea to unblock staging in the meantime 👍

PR #989, opened 2022-06-17
GRW-1212 / feat / Add possibility to redirect to any web page

simonauner:
## What

Add a new tag `<WebRedirect>` that allows us to redirect to market web (potentially any route) 

The idea is to use it like so:

```xml
<WebRedirect 
    path="/car-error" 
    when="createQuoteBundleResultTypename == QuoteBundleError"
></WebRedirect>
```

## Why

When we fail to create car quotes we could provide a lot more information to users than what we do today. The amount of information that could be provided won't fit into Embark. The idea is to redirect to a market-web page. It seems we do not have support for redirecting to any URL from Embark, so I added `<WebRedirect>` that does this. 

Adding the locale will be handled in `web-onboarding`:

```tsx
                <EmbarkProvider
                  externalRedirects={{
                    ...
                    Web: (route: string) => {
                      
                      // In case someone mistakenly adds a leading slash to the path
                      const r = route.replace(/^\//)
                      
                      history.push(`/${pathLocale}/${r}`)
                    }
                  }}
                  ...
                >
```

Jira ticket: [GRW-1212](https://hedvig.atlassian.net/browse/GRW-1212)


gustaveen (commented):
Nice! 👌 Would this work if you're in embark `/se-en`, then you would end up on `/se/car-error` right? So could we somehow pass the locale to the path?

simonauner:
> 

The implementation of `externalRedirects::Web` in `web-onboarding` would add the current locale. So if you're in Embark on `/se-en/new-member/car/cart` and get the error, 

```xml
<WebRedirect 
    path="/car-error" 
    when="createQuoteBundleResultTypename == QuoteBundleError"
></WebRedirect>
```

would redirect to `/se-en/car-error`.

PR #991, opened 2022-06-17
The sed command doesn't work on GH actions

simonauner:
Remove updating CHANGELOG in the `publish-npm` build config - `sed` works different on Linux and Mac, and I don't have the time to troubleshoot it.


guilhermespopolin (approved):

guilhermespopolin:
Would be the case to also remove `CHANGELOG.md` file?

simonauner:
> Would be the case to also remove `CHANGELOG.md` file?

It hurts me to say, but yes :(

simonauner:
> > Would be the case to also remove `CHANGELOG.md` file?
> 
> It hurts me to say, but yes :(

OR we actually manually update up 🙃

PR #994, opened 2022-06-20
Grw 1212/fix/translate redirect path

simonauner:


guilhermespopolin (dismissed):
That's a tricky one 😅 
As you've mentioned, the issue is that `useGoto` doesn't respect the order specified in the flow for _external redirects_.
One [_gambiarra_](https://www.urbandictionary.com/define.php?term=Gambiarra) solution would be to use _internal redirects_ with `<Redirect>` tag since it seems that for them the order you [specified in the flow is respected](https://github.com/HedvigInsurance/embark/blob/master/src/Utils/ExpressionsUtil/useGoTo.ts#L30). Here's is the ugly solution that it came into my mind:

Screenshot 2022-06-20 at 15 52 35

I know, it sucks. Just couldn't avoid the Brazilian spirit of _gambiarra_ that lives inside me.

simonauner:
> ...
> I know, it sucks. Just couldn't avoid the Brazilian spirit of _gambiarra_ that lives inside me.

Ah, ok, nice - then I could skip moving the if statements around and just fix it with a content update adding the internal redirect as you have in your image?

guilhermespopolin:
> specified in the flow is respected

Yes. It's not guaranteed that it will work tho; we'd need to test it. And the names I've chosen in the image are not the best.
Maybe we could rename `Offer` to `ErrorRedirectCheck`; `specificError` could be named `webRedirect` and `genericError` could be named `offerRedirect`.

simonauner:
> > specified in the flow is respected
> 
> Yes. It's not guaranteed that it will work tho; we'd need to test it. And the names I've chosen in the image are not the best. Maybe we could rename `Offer` to `ErrorRedirectCheck`; `specificError` could be named `webRedirect` and `genericError` could be named `offerRedirect`.

Ok, let's try it! I dropped the commit that changed the order of if-statements in `useGoTo` 👌

guilhermespopolin (approved):

PR #1000, opened 2022-06-29
GRW-1193 / Fix / Update external collection id references

robinandeer:
These were changed in the Insurely API back in 2020.

The next step is to request the provider status from the V2 endpoint.

Currently the API in staging responds with the lower case IDs so it breaks the onboarding.


dendoesit (approved):
👍

PR #1007, opened 2022-07-19
 added a variable when provider is not loading

dendoesit:


robinandeer (commented):
Same comment here about using a less generic variable name -- otherwise great!

[acceptit](https://app.graphite.dev/macros/HedvigInsurance)

guilhermespopolin (approved):

guilhermespopolin (approved):

PR #1014, opened 2022-08-04
chore(onboarding-NO): include support for student insurance

guilhermespopolin:
## What

Update `onboarding-NO` with the content of `onboarding-NO-v2` adding support for _student insurance_ to it.
After getting this released and have updated `onboarding` so it uses `onboarding-NO` instead of `onboarding-NO-v2` I can finally remove `onboarding-NO-v2` leaving one single flow for NO.
 
## Why

So we get back with only one embark flow in NO

[GRW-1272](https://hedvig.atlassian.net/browse/GRW-1272)


robinandeer (approved):
hehe interesting approach, I would've just keep going with v2 😅 -- but should work and will be cleaner

guilhermespopolin:
> hehe interesting approach, I would've just keep going with v2 😅 -- but should work and will be cleaner

Yeah, that would be the easiest way. Maybe I should have chosen that way 😅 
Let's leave it for the next time

PR #1015, opened 2022-08-09
feat(GRW-1221):  no/dk house cross-sell 

guilhermespopolin:
## What

Adds NO/DK onboarding flows that supports _House Cross-selling_

## Why

NO & DK embark flows needs to be changed so they can support House Cross-selling. Currently we have one mutation per combination of promise. The idea is to updated it so we can create quotes for all promises whenever is possible, which helps to enable House Cross-selling.

## Ticket
[GRW-1221](https://hedvig.atlassian.net/browse/GRW-1221)


guilhermespopolin:
For the newcomers @alebedev  and @KarnellSchultz. As you can see reviewing embark flows by checking the files' content is kind of impossible 😅 So I guess I'm not laying while saying that even though we create PR for them as well, at the end the reviewing process is basically just trust in the author 😅 

In any case, if you guys want I can walk you thro them. So just let me know.

simonauner (approved):
"LGTM"


 but have no idea what's going on :shipit:

PR #1018, opened 2022-08-26
Update to SE-onboarding-car

simonauner:
So it seems that the circular flow in https://github.com/HedvigInsurance/embark/pull/1017 didn't work. See how it looked below:

image

@robinandeer suggest just using a "Continue" button to take the user back to the `personalNumber` screen:

Screenshot 2022-08-26 at 10 04 54

image

image

Let's try it and see if it works better 🤞


robinandeer (approved):
[cheers](https://app.graphite.dev/macros/HedvigInsurance)

